PR #7, opened 2023-03-06
Upgrade go to 1.19 and related dependencies

ezio-auditore:
Upgrade list

- controller-tools
- operator-sdk
- gingko
- gomega
- operator-framework
- prometheus
- client-go
- controller-runtime
- k8s api

Other changes

- Promethues template
- k8sservicemonitor template
- golang ci lint version upgrade
- removed deprecated `ioutil ` package with `os` package
The above is required since after the prometheus package upgrade the spec has changed


.github/workflows/golangci-lint.yaml
```diff
@@ -18,5 +18,5 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.46.2
+          version: v1.50.0
```
bindrad, line 21: Why are we using v1.50.0?
latest version available is v1.51.2(latest)

Makefile
```diff
@@ -48,9 +48,9 @@ OPM ?= $(LOCALBIN)/opm
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v3.9.1
-CONTROLLER_TOOLS_VERSION ?= v0.8.0
+CONTROLLER_TOOLS_VERSION ?= v0.9.2
 ENVTEST_VERSION ?= latest
-OPERATOR_SDK_VERSION ?= v1.21.0
+OPERATOR_SDK_VERSION ?= v1.25.4
```
bindrad, line 53: Any specific reason to use v1.25.4?
v1.27.0 is available

bindrad: v1.25.0 and above supports multi-arch build support so why don't we add a target as suggested in version upgrade docs?

go.mod
```diff
@@ -50,22 +50,22 @@ require (
...
 	k8s.io/component-base v0.21.3 // indirect
-	k8s.io/klog/v2 v2.60.1 // indirect
+	k8s.io/klog/v2 v2.90.0 // indirect
 	k8s.io/kube-openapi v0.0.0-20220328201542-3ee0da9b0b42 // indirect
-	k8s.io/utils v0.0.0-20220210201930-3a6ce19ff2f9 // indirect
-	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
+	k8s.io/utils v0.0.0-20230202215443-34013725500c // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
```
bindrad, line 71: Shouldn't we remove modules from replace section as we are upgrading our modules to a newer version?

bindrad (changes requested):

.github/workflows/golangci-lint.yaml
ezio-auditore: Done

go.mod
ezio-auditore: Done

Makefile
ezio-auditore: v1.25.4 is recommended to be installed according to the docs
https://docs.okd.io/4.12/operators/operator_sdk/golang/osdk-golang-updating-projects.html

leelavg (approved):
- with the changes happened to prom, what's the lowest prom-op and ocp that we require?

PR #9, opened 2023-03-06
Initialize ManagedFusionOffering controller

bindrad:


leelavg:
how many commits in this PR should be reviewed?

controllers/managedfusionoffering_controller.go
```diff
@@ -46,18 +50,33 @@ type ManagedFusionOfferingReconciler struct {
...
-	_ = r.Log.WithValues("managedfusionoffering", req.NamespacedName)
+	log := r.Log.WithValues("managedfusionoffering", req.NamespacedName)
+	log.Info("Starting reconcile for ManangedFusionOffering")
 
-	// TODO(user): your logic here
+	managedFusionOfferingList := &v1alpha1.ManagedFusionOfferingList{}
+	if err := r.Client.List(ctx, managedFusionOfferingList); err != nil {
+		return ctrl.Result{}, err
+	}
+	if len(managedFusionOfferingList.Items) == 0 {
```
leelavg, line 60: am I correct in assuming that for each offering we'll have one offering CR?

controllers/managedfusionoffering_controller.go
```diff
@@ -46,18 +50,33 @@ type ManagedFusionOfferingReconciler struct {
...
+		r.Log.Info("no managedFusionOffering CR found")
+		return ctrl.Result{}, nil
+	}
+	for i := range managedFusionOfferingList.Items {
+		r.managedFusionOffering = &managedFusionOfferingList.Items[i]
+		switch r.managedFusionOffering.Spec.Kind {
+		case v1alpha1.KindDataFoundation:
+		case v1alpha1.KindDataFoundationClient:
```
leelavg, line 68: - the same format should be used in each reconcile function or we'll have a separate reconcile function for each offering?
- IOW, where'll the branch happen, in this switch statement itself or a switch in every function? I believe the former will be good.

bindrad:
> how many commits in this PR should be reviewed (only last 2)?

@leelavg Yes, only the last 2 commits should be reviewed for now

controllers/managedfusionoffering_controller.go
bindrad: Yes, correct

controllers/managedfusionoffering_controller.go
bindrad: For now, we will have a switch in Reconcile function based on the offering kind
Each offering kind will have its own initReconciler and reconcilePhases function
Example: initReconcilerForODF and reconcilePhasesForODF will be implemented for ODF offering

Later, Ohad will introduce a plugin system where any offering can plug into this controller.

controllers/managedfusionoffering_controller.go
bindrad: Just for clarity, assume we are supporting offering X and Y, both offerings can co-exist on the same cluster.
Then there will be two MFO CRs on the cluster, one for offering X and another for Y.

controllers/managedfusionoffering_controller.go
```diff
@@ -0,0 +1,82 @@
...
+	Scheme *runtime.Scheme
+
+	managedFusionOffering *v1alpha1.ManagedFusionOffering
+}
+
+//+kubebuilder:rbac:groups=misf.ibm.com,resources=managedfusionofferings,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=misf.ibm.com,resources=managedfusionofferings/status,verbs=get;update;patch
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
```
nb-ohad, line 44: This is a very generic comment which does not add a lot. Whoever is contributing code to this repo knows what a reconcile is.

controllers/managedfusionoffering_controller.go
```diff
@@ -0,0 +1,82 @@
...
+// TODO(user): Modify the Reconcile function to compare the state specified by
+// the ManagedFusionOffering object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.6.4/pkg/reconcile
```
nb-ohad, line 51: Remove that as well

controllers/managedfusionoffering_controller.go
```diff
@@ -0,0 +1,82 @@
...
+func (r *ManagedFusionOfferingReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := r.Log.WithValues("managedfusionoffering", req.NamespacedName)
```
nb-ohad, line 53: Why is the ket for `r.NamespacedName` in the log is `managedfusionoffering`?

```suggestion
	log := r.Log.WithValues("req.Namespace", req.Namespace, "req.Name", req.Name)
```

controllers/managedfusionoffering_controller.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	for i := range managedFusionOfferingList.Items {
+		r.managedFusionOffering = &managedFusionOfferingList.Items[i]
+		switch r.managedFusionOffering.Spec.Kind {
+		case v1alpha1.KindDataFoundation:
+		case v1alpha1.KindDataFoundationClient:
+		default:
+			return ctrl.Result{}, fmt.Errorf("Unknown managedFusionOffering Kind: %v", r.managedFusionOffering.Spec.Kind)
+		}
+	}
```
nb-ohad, line 72: Why are we listing all managed offering CRs? just load the one that the request refer to

main.go
```diff
@@ -37,6 +37,7 @@ import (
 	operators "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	promv1a1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1alpha1"
+	misfv1alpha1 "github.com/red-hat-storage/managed-fusion-agent/api/v1alpha1"
```
nb-ohad, line 40:
```suggestion
	misfv1a1 "github.com/red-hat-storage/managed-fusion-agent/api/v1alpha1"
```

controllers/managedfusionoffering_controller.go
```diff
@@ -0,0 +1,82 @@
...
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ManagedFusionOfferingReconciler) SetupWithManager(mgr ctrl.Manager, ctrlOptions *controller.Options) error {
```
nb-ohad, line 78: Can we please move the `SetupWithManager` above `reconcile`?

nb-ohad (changes requested):

controllers/managedfusionoffering_controller.go
bindrad: `2023-03-09T17:09:25.272+0530    INFO    controllers.ManagedFusionOffering       starting reconcile for ManangedFusionOffering   {"managedfusionoffering": "default/managedfusionoffering-sample"}`

bindrad:
> Why is main() inside the reconciler file and not in the main file? We cannot have 2 different main() functions for the different controllers. At least that is what I remember

@nb-ohad  Where are we doing this in the current PR?
main function is in main.go file and there is only a single main function

controllers/managedfusionoffering_controller.go

main.go
bindrad: Updated

controllers/managedfusionoffering_controller.go

controllers/managedfusionoffering_controller.go

controllers/managedfusionoffering_controller.go

controllers/managedfusionoffering_controller.go
```diff
@@ -0,0 +1,82 @@
...
+	if err := r.get(r.managedFusionOffering); err != nil {
+		return ctrl.Result{}, fmt.Errorf("Unable to get managedFusionOffering: %v", err)
+	}
+
+	switch r.managedFusionOffering.Spec.Kind {
+	case v1alpha1.KindDataFoundation:
+	case v1alpha1.KindDataFoundationClient:
+	default:
+		return ctrl.Result{}, fmt.Errorf("Unknown managedFusionOffering Kind: %v", r.managedFusionOffering.Spec.Kind)
+	}
```
nb-ohad, line 66: This check and the enum itself do not make sense.
The plan is to introduce offerings as plugins, so any value should be ok as long as we can find a plugin for it and I don't see how can we do it in compile time. as such we should assume the field a a plain old string and no validation for now

nb-ohad (changes requested):

controllers/managedfusionoffering_controller.go
bindrad:
```suggestion
	if result, err := r.reconcileDF(); err != nil {
		r.Log.Error(err, "An error was encountered during reconcile of DF offering")
		return ctrl.Result{}, err
	} else {
		return result, nil
	}
```

So for now(without a plugin system), can I assume that a function called reconcileDF function will be called in Reconcile function?

nb-ohad: First, For this PR nothing should be called as this is not a DF PR.
Second, we already discussed in the morning should abstract it behind a `pluginReconcile(...)` call the internally calls your `reconcileDF`. Of course, `pluginReconcile` is not a method of the reconciler so you will need to pass the reconciler as a parameter. Something like:

```
result, err := pluginReconcile(r, ...);  
```

nb-ohad:
@bindrad I might got confused with file boundaries when doing the review. Sorry for that

controllers/managedfusionoffering_controller.go
```diff
@@ -0,0 +1,89 @@
...
+	ctx                   context.Context
+	managedFusionOffering *v1alpha1.ManagedFusionOffering
+}
+
+//+kubebuilder:rbac:groups=misf.ibm.com,resources=managedfusionofferings,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=misf.ibm.com,resources=managedfusionofferings/status,verbs=get;update;patch
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ManagedFusionOfferingReconciler) SetupWithManager(mgr ctrl.Manager, ctrlOptions *controller.Options) error {
+	managedFusionOfferingController := ctrl.NewControllerManagedBy(mgr).
```
nb-ohad, line 47: This is not a controller, it is a controller builder. The controller is returned only after you call the complete method.
Also, no need to mention managedFusion as this is building a single controller under the context of this `SetupWithManager` call. 

. 
```suggestion
	controllerBuilder  := ctrl.NewControllerManagedBy(mgr).
```

controllers/managedfusionoffering_controller.go
```diff
@@ -0,0 +1,89 @@
...
+		For(&v1alpha1.ManagedFusionOffering{})
+
+	pluginSetupWatches(r, managedFusionOfferingController)
```
nb-ohad, line 50: Do we really need the reconciler here? What kind of information do you think the plugin will need from the reconciler?

controllers/managedfusionoffering_controller.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (r *ManagedFusionOfferingReconciler) initReconciler(ctx context.Context, req ctrl.Request) {
+	r.ctx = ctx
+
+	r.managedFusionOffering = &v1alpha1.ManagedFusionOffering{}
+	r.managedFusionOffering.Name = req.Name
+	r.managedFusionOffering.Namespace = req.Namespace
+}
+
+func pluginReconcile(r *ManagedFusionOfferingReconciler) (ctrl.Result, error) {
```
nb-ohad, line 80: Add a comment saying that this function is a placeholder for offering plugin integration

controllers/managedfusionoffering_controller.go
```diff
@@ -0,0 +1,89 @@
...
+	return ctrl.Result{}, nil
+}
+
+func pluginSetupWatches(r *ManagedFusionOfferingReconciler, controller *builder.Builder) {}
```
nb-ohad, line 84: Add a comment saying that this function is a placeholder for offering plugin integration

main.go
```diff
@@ -133,3 +147,5 @@ func getUnrestrictedClient() client.Client {
 	}
 	return k8sClient
 }
+
+func pluginAddToScheme(scheme *runtime.Scheme) {}
```
nb-ohad, line 151: Add a comment saying that this function is a placeholder for offering plugin integration

main.go
```diff
@@ -69,6 +70,10 @@ func addAllSchemes(scheme *runtime.Scheme) {
 	utilruntime.Must(openshiftv1.AddToScheme(scheme))
 
 	utilruntime.Must(configv1.AddToScheme(scheme))
+
+	utilruntime.Must(misfv1a1.AddToScheme(scheme))
+
+	pluginAddToScheme(scheme)
 	// +kubebuilder:scaffold:scheme
```
nb-ohad, line 77: Move the scaffolding comment above the plugin code and leave an empty line

nb-ohad (changes requested):

controllers/managedfusionoffering_controller.go
bindrad: Updated

controllers/managedfusionoffering_controller.go

main.go
bindrad: Updated

controllers/managedfusionoffering_controller.go
bindrad: No, we won't need reconciler to setupWatches
I don't think we will need any reconciler fields initialized in main.go
Updated the code accordingly

nb-ohad (dismissed):

PR #13, opened 2023-03-14
Added CephIngressNetworkPolicy to plugin reconcile funcs

ezio-auditore:


controllers/datafoundation.go
```diff
@@ -114,3 +120,21 @@ func (r *dataFoundationReconciler) reconcileOnboardingValidationSecret() error {
 	}
 	return nil
 }
+
+func (r *dataFoundationReconciler) reconcileCephIngressNetworkPolicy() error {
+	cephIngressNetworkPolicy := &netv1.NetworkPolicy{}
+	cephIngressNetworkPolicy.Name = "ceph-ingress-rule"
+	cephIngressNetworkPolicy.Namespace = r.managedFusionOffering.Namespace
```
nb-ohad, line 127: The current convention in the codebase it to initialize the identity of the reconciled resources in init function. Please follow the guideline

controllers/datafoundation.go
```diff
@@ -114,3 +120,21 @@ func (r *dataFoundationReconciler) reconcileOnboardingValidationSecret() error {
...
+	_, err := ctrl.CreateOrUpdate(r.Ctx, r.Client, cephIngressNetworkPolicy, func() error {
+		if err := r.own(cephIngressNetworkPolicy); err != nil {
+			return err
+		}
+		desired := templates.CephNetworkPolicyTemplate.DeepCopy()
```
nb-ohad, line 132: I don't see a template as part of this PR. 
But the templates package is the general template package and not the datafoundation one which is a mistake.

controllers/managedfusionoffering_controller.go
```diff
@@ -56,6 +56,7 @@ type ManagedFusionOfferingReconciler struct {
 //+kubebuilder:rbac:groups=misf.ibm.com,resources={managedfusionofferings,managedfusionofferings/finalizers},verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=misf.ibm.com,resources=managedfusionofferings/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups="operators.coreos.com",resources={subscriptions,operatorgroups,catalogsources},verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="network.openshift.io",resources=egressnetworkpolicies,verbs=create;get;list;watch;update
```
nb-ohad, line 59: Isn't this change needs an ingress policy RBAC and not an egress policy RBAC?
Also if we want to have RBAC for an offering specific resource we should move this line to the datafoundation package

nb-ohad (changes requested):

controllers/datafoundation.go
ezio-auditore: Updated

controllers/datafoundation.go
ezio-auditore: Corrected.

controllers/managedfusionoffering_controller.go
ezio-auditore: Yes, updated accordingly

nb-ohad (approved):

PR #16, opened 2023-03-14
Add reconcile onboarding validation secret function

bindrad:
This PR introduces the reconcileOnboardingValidationSecret function that will create a secret which will have a public key that will be used to onboard the consumer clusters.

### DIFF
--- a/controllers/datafoundation.go
+++ b/controllers/datafoundation.go
@@ -1,0 +1,116 @@
+package controllers
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	v1alpha1 "github.com/red-hat-storage/managed-fusion-agent/api/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+)
+
+type dataFoundationSpec struct {
+	usableCapacityInTiB     int
+	onboardingValidationKey string
+}
+
+type dataFoundationReconciler struct {
+	*ManagedFusionOfferingReconciler
+
+	spec                          dataFoundationSpec
+	onboardingValidationKeySecret *corev1.Secret
+}
+
+func dfSetupWatches(controllerBuilder *builder.Builder) {
+	controllerBuilder.
+		Owns(&corev1.Secret{})
+}
+
+func dfReconcile(offeringReconciler *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) error {
+	r := dataFoundationReconciler{}
+	r.initReconciler(offeringReconciler)
+
+	if err := r.parseSpec(offering); err != nil {
+		return err
+	}
+	if err := r.reconcileOnboardingValidationSecret(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (r *dataFoundationReconciler) initReconciler(offeringReconciler *ManagedFusionOfferingReconciler) {
+	r.ManagedFusionOfferingReconciler = offeringReconciler
+
+	r.onboardingValidationKeySecret = &corev1.Secret{}
+	r.onboardingValidationKeySecret.Name = "onboarding-ticket-key"
+	r.onboardingValidationKeySecret.Namespace = r.Namespace
+}
+
+func (r *dataFoundationReconciler) parseSpec(offering *v1alpha1.ManagedFusionOffering) error {
+	r.Log.Info("Parsing ManagedFusionOffering Data Foundation spec")
+
+	valid := true
+	var usableCapacityInTiB int
+	var err error
+	if usableCapacityInTiBAsString, found := offering.Spec.Config["usableCapacityInTiB"]; !found {
+		r.Log.Error(
+			fmt.Errorf("missing field: usableCapacityInTiB"),
+			"an error occurred while parsing ManagedFusionOffering Data Foundation spec",
+		)
+		valid = false
+	} else if usableCapacityInTiB, err = strconv.Atoi(usableCapacityInTiBAsString); err != nil {
+		r.Log.Error(
+			fmt.Errorf("error parsing usableCapacityInTib: %v", err),
+			"an error occurred while parsing ManagedFusionOffering Data Foundation spec",
+		)
+		valid = false
+	}
+
+	var onboardingValidationKeyAsString string
+	var found bool
+	if onboardingValidationKeyAsString, found = offering.Spec.Config["onboardingValidationKey"]; !found {
+		r.Log.Error(
+			fmt.Errorf("missing field: onboardingValidationKey"),
+			"an error occurred while parsing ManagedFusionOffering Data Foundation spec",
+		)
+		valid = false
+	}
+
+	if !valid {
+		r.Log.Info("parsing ManagedFusionOffering Data Foundation spec failed")
+		return fmt.Errorf("invalid ManagedFusionOffering Data Foundation spec")
+	}
+	r.Log.Info("parsing ManagedFusionOffering Data Foundation spec completed successfuly")
+
+	r.spec = dataFoundationSpec{
+		usableCapacityInTiB:     usableCapacityInTiB,
+		onboardingValidationKey: onboardingValidationKeyAsString,
+	}
+	return nil
+}
+
+func (r *dataFoundationReconciler) reconcileOnboardingValidationSecret() error {
+	r.Log.Info("Reconciling onboardingValidationKey secret")
+
+	_, err := ctrl.CreateOrUpdate(r.Ctx, r.Client, r.onboardingValidationKeySecret, func() error {
+		if err := r.own(r.onboardingValidationKeySecret); err != nil {
+			return err
+		}
+		onboardingValidationData := fmt.Sprintf(
+			"-----BEGIN PUBLIC KEY-----\n%s\n-----END PUBLIC KEY-----",
+			strings.TrimSpace(r.spec.onboardingValidationKey),
+		)
+		r.onboardingValidationKeySecret.Data = map[string][]byte{
+			"key": []byte(onboardingValidationData),
+		}
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to update onboardingValidationKeySecret: %v", err)
+	}
+	return nil
+}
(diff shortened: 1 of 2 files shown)


leelavg (approved):
- the plugin functions seems to be in-directions atm 😅, will see how that works

controllers/managedfusionoffering_controller.go
```diff
@@ -79,11 +82,55 @@ func (r *ManagedFusionOfferingReconciler) initReconciler(ctx context.Context, re
...
+
 // This function is a placeholder for offering plugin integration
-func pluginSetupWatches(controllerBuilder *builder.Builder) {}
+func pluginSetupWatches(controllerBuilder *builder.Builder) {
+	controllerBuilder.
+		Owns(&corev1.Secret{})
+}
+
+func (r *ManagedFusionOfferingReconciler) own(resource metav1.Object) error {
+	// Ensure managedOCS ownership on a resource
```
ezio-auditore, line 128: To be changed to managedFusion

leelavg (dismissed):

ezio-auditore (dismissed):

controllers/managedfusionoffering_controller.go
```diff
@@ -85,11 +88,49 @@ func pluginReconcile(r *ManagedFusionOfferingReconciler) (ctrl.Result, error) {
 	if _, found := r.managedFusionOffering.Spec.Config["onboardingValidationKey"]; !found {
 		return ctrl.Result{}, fmt.Errorf("%s CR does not contain onboardingValidationKey entry", r.managedFusionOffering.Name)
 	}
+	if err := reconcileOnboardingValidationSecret(r); err != nil {
```
nb-ohad, line 91: Why not ?
```suggestion 
	if err := r.reconcileOnboardingValidationSecret(); err != nil {
```
This will introduce less changes when we copy the code from the old deployer

controllers/managedfusionoffering_controller.go
```diff
@@ -85,11 +88,49 @@ func pluginReconcile(r *ManagedFusionOfferingReconciler) (ctrl.Result, error) {
...
+		return ctrl.Result{}, err
+	}
 	return ctrl.Result{}, nil
 }
 
+func reconcileOnboardingValidationSecret(r *ManagedFusionOfferingReconciler) error {
+	r.Log.Info("Reconciling onboardingValidationKeySecret")
+	onboardingValidationKeySecret := &corev1.Secret{}
```
nb-ohad, line 99: Please add a space after initial log

controllers/managedfusionoffering_controller.go
```diff
@@ -85,11 +88,49 @@ func pluginReconcile(r *ManagedFusionOfferingReconciler) (ctrl.Result, error) {
...
+		Owns(&corev1.Secret{})
+}
+
+func (r *ManagedFusionOfferingReconciler) own(resource metav1.Object) error {
+	// Ensure managedFusionOffering ownership on a resource
+	if err := ctrl.SetControllerReference(r.managedFusionOffering, resource, r.Scheme); err != nil {
+		return err
+	}
+	return nil
+}
```
nb-ohad, line 133: I think was already introduced in a previous PR

controllers/managedfusionoffering_controller.go
```diff
@@ -85,11 +88,49 @@ func pluginReconcile(r *ManagedFusionOfferingReconciler) (ctrl.Result, error) {
...
+	onboardingValidationKeySecret := &corev1.Secret{}
+	onboardingValidationKeySecret.Name = "onboarding-ticket-key"
+	onboardingValidationKeySecret.Namespace = r.managedFusionOffering.Namespace
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, onboardingValidationKeySecret, func() error {
+		if err := r.own(onboardingValidationKeySecret); err != nil {
+			return err
+		}
+		onboardingValidationData := fmt.Sprintf(
+			"-----BEGIN PUBLIC KEY-----\n%s\n-----END PUBLIC KEY-----",
+			strings.TrimSpace(r.managedFusionOffering.Spec.Config["onboardingValidationKey"]),
```
nb-ohad, line 108: I consider reaching into the unstructured config a bad practice. My preference will be to have a structured struct that we set up at the start of the reconciliation (with the validation) and use it throughout the reconciliation. 

This also brings the thought of using a marshaled YAML document as an offering config instead of a string-to-string map.

nb-ohad (changes requested):

controllers/managedfusionoffering_controller.go
bindrad: Updated

controllers/managedfusionoffering_controller.go
bindrad: > My preference will be to have a structured struct that we set up at the start of the reconciliation (with the validation) and use it throughout the reconciliation.

Updated

> This also brings the thought of using a marshaled YAML document as an offering config instead of a string-to-string map.

As discussed, we will see if this needs to be done in future.

controllers/managedfusionoffering_controller.go
nb-ohad: I believe so. why do you think we will not?

bindrad: I was not able to make it work

nb-ohad: You can embed it, 
see here: https://go.dev/play/p/oY-ZBOU99JB

bindrad: Yes, this will work but in our case reconciler will be coming from another package so you cannot do the same thing.

controllers/datafoundation.go
```diff
@@ -0,0 +1,113 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+)
+
+type dataFoundationSpec struct {
+	usableCapacityInTiB     int
+	onboardingValidationKey string
+}
+type dataFoundationReconciler struct {
+	*ManagedFusionOfferingReconciler
```
nb-ohad, line 18:
```suggestion
	*ManagedFusionOfferingReconciler
	
```

controllers/datafoundation.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func dfSetupWatches(controllerBuilder *builder.Builder) {
+	controllerBuilder.
+		Owns(&corev1.Secret{})
+}
+
+func dfReconcile(managedFusionOfferingReconciler *ManagedFusionOfferingReconciler) error {
+	var r *dataFoundationReconciler
+	r.ManagedFusionOfferingReconciler = managedFusionOfferingReconciler
```
nb-ohad, line 30: This will throw a nil access error. You need to create a concrete `dataFoundationReconciler` before you can assign to any field including the `ManagedFusionOfferingReconciler ` embedded field. Also I see no reason to use a pointer here in the first place.

  ```suggestion
	r := dataFoundationReconciler{
		ManagedFusionOfferingReconciler: ManagedFusionOfferingReconciler
	}
```

But there is even a better way. You have an `r.initReconciler` function that you call just a couple of lines below. Why not use it to set up the embedded field, the same as you do for other fields.

Then this line can be as simple as 
```suggestion
	r := dataFoundationReconciler{
```

And the first line in `initReconcier` will be
```
	r.ManagedFusionOfferingReconciler = offeringReconciler
```

Assuming `offeringReconciler` will be the name of the input argument

controllers/datafoundation.go
```diff
@@ -0,0 +1,113 @@
...
+
+	r.initDataFoundationReconciler()
```
nb-ohad, line 32: No need to specify `DataFoundation` on every method name of the `DataFoundationReconciler`
```suggestion
	r.initReconciler()
```

controllers/datafoundation.go
```diff
@@ -0,0 +1,113 @@
...
+
+	if err := r.parseDataFoundationSpec(); err != nil {
```
nb-ohad, line 34: Shouldn't this function get the ManagedFusionOffering.Spec.Config map as input?
Also name can be shorter and still meaningful under this context
```suggestion
	if err := r.parseSpec(); err != nil {
```

controllers/datafoundation.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	if err := r.reconcileOnboardingValidationSecret(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (r *dataFoundationReconciler) initDataFoundationReconciler() {
+	r.onboardingValidationKeySecret = &corev1.Secret{}
```
nb-ohad, line 45: As discussed in a prev comment

```suggestion
func (r *dataFoundationReconciler) initDataFoundationReconciler(offeringReconciler  *ManagedFusionOfferingReconciler) {
	r.ManagedFusionOfferingReconciler = offeringReconciler 

	r.onboardingValidationKeySecret = &corev1.Secret{}
```

controllers/datafoundation.go
```diff
@@ -0,0 +1,113 @@
...
+	r.onboardingValidationKeySecret.Name = "onboarding-ticket-key"
+	r.onboardingValidationKeySecret.Namespace = r.Namespace
+}
+
+func (r *dataFoundationReconciler) parseDataFoundationSpec() error {
```
nb-ohad, line 50: No need to mention `DataFoundation` for every `dataFoundationReconciler` method
```suggestion
func (r *dataFoundationReconciler) parseSpec() error {
```

controllers/managedfusionoffering_controller.go
```diff
@@ -46,9 +45,9 @@ type ManagedFusionOfferingReconciler struct {
 	Log    logr.Logger
 	Scheme *runtime.Scheme
 
-	ctx                   context.Context
-	namespace             string
-	managedFusionOffering *v1alpha1.ManagedFusionOffering
+	Ctx                   context.Context
+	Namespace             string
+	ManagedFusionOffering *v1alpha1.ManagedFusionOffering
```
nb-ohad, line 50: No reason to make this public. just pass it as a private param to the plugin reconcile code.
We assume the Embbeding-reconcilers have no direct knowledge of the specific of the embedded reconciler besides some interface that will include commons such as Ctx and maybe Namespace.

nb-ohad (changes requested):

nb-ohad (changes requested):

nb-ohad (approved):

PR #20, opened 2023-03-14
Add reconcile CSVs function

bindrad:
This PR has changes to reconcile OCS and MCG operator's CSV

### DIFF
--- a/controllers/managedfusionoffering_controller.go
+++ b/controllers/managedfusionoffering_controller.go
@@ -19,14 +19,30 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/go-logr/logr"
+	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	v1alpha1 "github.com/red-hat-storage/managed-fusion-agent/api/v1alpha1"
+	"github.com/red-hat-storage/managed-fusion-agent/utils"
+	"k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	controller "sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	managedFusionAnnotation = "misf.ibm.com/managedfusionoffering"
+	ocsOperatorName         = "ocs-operator"
+	mcgOperatorName         = "mcg-operator"
 )
 
 // ManagedFusionOfferingReconciler reconciles a ManagedFusionOffering object
@@ -85,13 +101,155 @@ func pluginReconcile(r *ManagedFusionOfferingReconciler) (ctrl.Result, error) {
 	if _, found := r.managedFusionOffering.Spec.Config["onboardingValidationKey"]; !found {
 		return ctrl.Result{}, fmt.Errorf("%s CR does not contain onboardingValidationKey entry", r.managedFusionOffering.Name)
 	}
+	if err := reconcileCSV(r); err != nil {
+		return ctrl.Result{}, err
+	}
 	return ctrl.Result{}, nil
 }
 
+func reconcileCSV(r *ManagedFusionOfferingReconciler) error {
+	r.Log.Info("Reconciling CSVs")
+	csvList := opv1a1.ClusterServiceVersionList{}
+	if err := r.list(&csvList); err != nil {
+		return fmt.Errorf("unable to list csv resources: %v", err)
+	}
+
+	for index := range csvList.Items {
+		csv := &csvList.Items[index]
+		if strings.HasPrefix(csv.Name, ocsOperatorName) {
+			if err := updateOCSCSV(r, csv); err != nil {
+				return fmt.Errorf("Failed to update OCS CSV: %v", err)
+			}
+		} else if strings.HasPrefix(csv.Name, mcgOperatorName) {
+			if err := updateMCGCSV(r, csv); err != nil {
+				return fmt.Errorf("Failed to update MCG CSV: %v", err)
+			}
+		}
+	}
+	return nil
+}
+
+func updateOCSCSV(r *ManagedFusionOfferingReconciler, csv *opv1a1.ClusterServiceVersion) error {
+	isChanged := false
+	if _, found := csv.Annotations[managedFusionAnnotation]; !found {
+		r.addManagedFusionOfferingAnnotation(csv, managedFusionAnnotation)
+		isChanged = true
+	}
+	deployments := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs
+	for i := range deployments {
+		containers := deployments[i].Spec.Template.Spec.Containers
+		for j := range containers {
+			switch container := &containers[j]; container.Name {
+			case "ocs-operator":
+				resources := utils.GetResourceRequirements("ocs-operator")
+				if !equality.Semantic.DeepEqual(container.Resources, resources) {
+					container.Resources = resources
+					isChanged = true
+				}
+			case "rook-ceph-operator":
+				resources := utils.GetResourceRequirements("rook-ceph-operator")
+				if !equality.Semantic.DeepEqual(container.Resources, resources) {
+					container.Resources = resources
+					isChanged = true
+				}
+			case "ocs-metrics-exporter":
+				resources := utils.GetResourceRequirements("ocs-metrics-exporter")
+				if !equality.Semantic.DeepEqual(container.Resources, resources) {
+					container.Resources = resources
+					isChanged = true
+				}
+			default:
+				r.Log.V(-1).Info("Could not find resource requirement", "Resource", container.Name)
+			}
+		}
+	}
+	if isChanged {
+		if err := r.update(csv); err != nil {
+			return fmt.Errorf("Failed to update OCS CSV: %v", err)
+		}
+	}
+	return nil
+}
+
+func updateMCGCSV(r *ManagedFusionOfferingReconciler, csv *opv1a1.ClusterServiceVersion) error {
+	isChanged := false
+	if _, found := csv.Annotations[managedFusionAnnotation]; !found {
+		r.addManagedFusionOfferingAnnotation(csv, managedFusionAnnotation)
+		isChanged = true
+	}
+	mcgDeployments := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs
+	for i := range mcgDeployments {
+		deployment := &mcgDeployments[i]
+		// Disable noobaa operator by scaling down the replica of noobaa deploymnet
+		// in MCG Operator CSV.
+		if deployment.Name == "noobaa-operator" &&
+			(deployment.Spec.Replicas == nil || *deployment.Spec.Replicas > 0) {
+			zero := int32(0)
+			deployment.Spec.Replicas = &zero
+			isChanged = true
+		}
+	}
+	if isChanged {
+		if err := r.update(csv); err != nil {
+			return fmt.Errorf("Failed to update MCG CSV: %v", err)
+		}
+	}
+	return nil
+}
+
 // This function is a placeholder for offering plugin integration
-func pluginSetupWatches(controllerBuilder *builder.Builder) {}
+func pluginSetupWatches(controllerBuilder *builder.Builder) {
+	csvPredicates := builder.WithPredicates(
+		predicate.NewPredicateFuncs(
+			func(client client.Object) bool {
+				return strings.HasPrefix(client.GetName(), ocsOperatorName) ||
+					strings.HasPrefix(client.GetName(), mcgOperatorName)
+			},
+		),
+	)
+	enqueueManagedFusionOfferingRequest := handler.EnqueueRequestsFromMapFunc(
+		func(object client.Object) []reconcile.Request {
+			annotations := object.GetAnnotations()
+			if annotation, found := annotations[managedFusionAnnotation]; found {
+				namespacedName := strings.Split(annotation, "/")
+				return []reconcile.Request{{
+					NamespacedName: types.NamespacedName{
+						Namespace: namespacedName[0],
+						Name:      namespacedName[1],
+					},
+				}}
+			}
+			return []reconcile.Request{}
+		},
+	)
+	controllerBuilder.
+		Watches(
+			&source.Kind{Type: &opv1a1.ClusterServiceVersion{}},
+			enqueueManagedFusionOfferingRequest,
+			csvPredicates,
+		)
+}
+
+func (r *ManagedFusionOfferingReconciler) addManagedFusionOfferingAnnotation(obj metav1.Object, key string) {
+	annotations := obj.GetAnnotations()
+	if annotations == nil {
+		annotations = map[string]string{}
+		obj.SetAnnotations(annotations)
+	}
+	value := fmt.Sprintf("%s/%s", r.managedFusionOffering.Namespace, r.managedFusionOffering.Name)
+	annotations[key] = value
+}
+
+func (r *ManagedFusionOfferingReconciler) list(obj client.ObjectList) error {
+	listOptions := client.InNamespace(r.managedFusionOffering.Namespace)
+	return r.Client.List(r.ctx, obj, listOptions)
+}
 
 func (r *ManagedFusionOfferingReconciler) get(obj client.Object) error {
 	key := client.ObjectKeyFromObject(obj)
 	return r.Client.Get(r.ctx, key, obj)
+}
+
+func (r *ManagedFusionOfferingReconciler) update(obj client.Object) error {
+	return r.Client.Update(r.ctx, obj)
 }
--- a/controllers/managedfusionoffering_controller.go
+++ b/controllers/managedfusionoffering_controller.go
@@ -79,6 +79,12 @@ func (r *ManagedFusionOfferingReconciler) initReconciler(ctx context.Context, re
 
 // This function is a placeholder for offering plugin integration
 func pluginReconcile(r *ManagedFusionOfferingReconciler) (ctrl.Result, error) {
+	if _, found := r.managedFusionOffering.Spec.Config["usableCapacityInTiB"]; !found {
+		return ctrl.Result{}, fmt.Errorf("%s CR does not contain usableCapacityInTiB entry", r.managedFusionOffering.Name)
+	}
+	if _, found := r.managedFusionOffering.Spec.Config["onboardingValidationKey"]; !found {
+		return ctrl.Result{}, fmt.Errorf("%s CR does not contain onboardingValidationKey entry", r.managedFusionOffering.Name)
+	}
 	return ctrl.Result{}, nil
 }
 


controllers/managedfusionoffering_controller.go
```diff
@@ -181,18 +189,41 @@ func (r *ManagedFusionOfferingReconciler) reconcileSubscription() error {
...
 
+func (r *ManagedFusionOfferingReconciler) addManagedFusionOfferingAnnotation(obj metav1.Object, key string) {
+	annotations := obj.GetAnnotations()
+	if annotations == nil {
+		annotations = map[string]string{}
+		obj.SetAnnotations(annotations)
+	}
+	value := fmt.Sprintf("%s/%s", r.managedFusionOffering.Namespace, r.managedFusionOffering.Name)
+	annotations[key] = value
+}
```
nb-ohad, line 200: We have a util for adding annotations so why are we implementing this here?

controllers/managedfusionoffering_controller.go
```diff
@@ -181,18 +189,41 @@ func (r *ManagedFusionOfferingReconciler) reconcileSubscription() error {
...
 func (r *ManagedFusionOfferingReconciler) own(resource metav1.Object) error {
 	// Ensure ManangedFusionOffering CR ownership on a resource
 	return ctrl.SetControllerReference(r.managedFusionOffering, resource, r.Scheme)
 }
 
 // All the below functions are placeholder for offering plugin integration
 
+const (
+	ocsOperatorName = "ocs-operator"
+)
```
nb-ohad, line 225: We should not have an operator name in this global context

nb-ohad (changes requested):

controllers/managedfusionoffering_controller.go

controllers/managedfusionoffering_controller.go

controllers/managedfusionoffering_controller.go
```diff
@@ -249,8 +284,98 @@ func parseDataFoundationSpec(r *ManagedFusionOfferingReconciler) (dataFoundation
...
-func pluginSetupWatches(controllerBuilder *builder.Builder) {}
+func pluginSetupWatches(controllerBuilder *builder.Builder) {
+	csvPredicates := builder.WithPredicates(
+		predicate.NewPredicateFuncs(
+			func(client client.Object) bool {
+				return strings.HasPrefix(client.GetName(), ocsOperatorName)
+			},
+		),
+	)
+	enqueueManagedFusionOfferingRequest := handler.EnqueueRequestsFromMapFunc(
```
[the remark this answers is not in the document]

controllers/managedfusionoffering_controller.go
```diff
@@ -249,8 +284,98 @@ func parseDataFoundationSpec(r *ManagedFusionOfferingReconciler) (dataFoundation
...
+		r.addManagedFusionOfferingAnnotation(csv, managedFusionAnnotation)
+		isChanged = true
+	}
+	deployments := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs
+	for i := range deployments {
+		containers := deployments[i].Spec.Template.Spec.Containers
+		for j := range containers {
+			switch container := &containers[j]; container.Name {
+			case "ocs-operator":
+				resources := utils.GetResourceRequirements("ocs-operator")
```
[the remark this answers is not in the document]

controllers/datafoundation.go
```diff
@@ -5,10 +5,21 @@ import (
...
+	"k8s.io/apimachinery/pkg/api/equality"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	ocsOperatorName = "ocs-operator"
```
nb-ohad, line 22: Is that the operator name or the packageName?

controllers/datafoundation.go
```diff
@@ -19,32 +30,53 @@ type dataFoundationSpec struct {
...
+	}
 
 	return nil
 }
 
-func (r *dataFoundationReconciler) initReconciler(offeringReconciler *ManagedFusionOfferingReconciler) {
+func (r *dataFoundationReconciler) initReconciler(offeringReconciler *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) {
 	r.ManagedFusionOfferingReconciler = offeringReconciler
 
+	r.offering = offering
```
nb-ohad, line 78: If we only care about the Namespaced Name of the offering, let's just save that. or even just the name and namespace separately. And to think about it you already have access to the namespace via r.Namespace. So maybe it is better to make r.Namespace private and get it from the offering resource
 ```suggestion
	r.offeringNamespace = offering.Namespace
	r.offeringName = offeringName
	
```

File: controllers/datafoundation.go [hunk 7]
```diff
@@ -114,3 +146,68 @@ func (r *dataFoundationReconciler) reconcileOnboardingValidationSecret() error {
...
+			}
+			break
+		}
+	}
+	return nil
+}
+
+func (r *dataFoundationReconciler) updateOCSCSV(csv *opv1a1.ClusterServiceVersion) error {
+	isChanged := false
+	if _, found := csv.Annotations[managedFusionAnnotation]; !found {
```
nb-ohad, line 171: If we have a util, why do we need to do this check? 
Also, finding the annotation does not mean it has the correct value

controllers/managedfusionoffering_controller.go
```diff
@@ -180,11 +181,20 @@ func (r *ManagedFusionOfferingReconciler) reconcileSubscription() error {
 	return nil
 }
 
+func (r *ManagedFusionOfferingReconciler) list(obj client.ObjectList) error {
+	listOptions := client.InNamespace(r.managedFusionOffering.Namespace)
+	return r.Client.List(r.Ctx, obj, listOptions)
+}
```
nb-ohad, line 187: This does not make sense anymore. We are now a multi-namespace operator, as such, we need the ability to list across namespaces. In that case, let's do

```suggestion
func (r *ManagedFusionOfferingReconciler) list(obj client.ObjectList, ...client.ListOption) error {
	return r.Client.List(r.Ctx, obj, listOptions...)
}
```

controllers/datafoundation.go
```diff
@@ -114,3 +146,68 @@ func (r *dataFoundationReconciler) reconcileOnboardingValidationSecret() error {
...
+		utils.AddAnnotation(
+			csv,
+			managedFusionAnnotation,
```
nb-ohad, line 174: This is not local to the offering so it makes no sense to have access to it here. Maybe you want an operation like `r.addOwnerAnnotation(res client.Object)` added to the embedded reconciler

nb-ohad (changes requested):

controllers/managedfusionoffering_controller.go

controllers/datafoundation.go

controllers/managedfusionoffering_controller.go
```diff
@@ -209,7 +228,7 @@ func pluginSetupWatches(controllerBuilder *builder.Builder) {
 // This function is a placeholder for offering plugin integration
 func pluginGetDesiredOperatorGroupSpec(r *ManagedFusionOfferingReconciler) opv1.OperatorGroupSpec {
 	return opv1.OperatorGroupSpec{
-		TargetNamespaces: []string{r.Namespace},
+		TargetNamespaces: []string{r.namespace},
```
nb-ohad, line 231: I think the decision here needs the offering CR, not the reconciler. If you are in the need to acquire the namespace of the offering you should take it from the offering cr itself. 

Please remember that you will not have access to private fields of the ManagedFusionOfferingReconciler  when we move this file to the datafoundation package

controllers/managedfusionoffering_controller.go
```diff
@@ -180,11 +182,28 @@ func (r *ManagedFusionOfferingReconciler) reconcileSubscription() error {
...
+func (r *ManagedFusionOfferingReconciler) addOwnerAnnotation(res client.Object) {
+	utils.AddAnnotation(
+		res,
+		managedFusionAnnotation,
+		fmt.Sprintf("%s/%s", r.namespace, r.managedFusionOffering.Name),
+	)
+}
+
+func (r *ManagedFusionOfferingReconciler) list(obj client.ObjectList, listOptions ...client.ListOption) error {
+	// listOptions := client.InNamespace(r.managedFusionOffering.Namespace)
```
nb-ohad, line 194: Why do we need this comment?

controllers/datafoundation.go
```diff
@@ -114,3 +148,61 @@ func (r *dataFoundationReconciler) reconcileOnboardingValidationSecret() error {
...
+				r.Log.V(-1).Info("Could not find resource requirement", "Resource", container.Name)
+			}
+		}
+	}
+	if isChanged {
+		if err := r.update(csv); err != nil {
+			return fmt.Errorf("failed to update OCS CSV: %v", err)
+		}
+	}
+	return nil
```
nb-ohad, line 207: Your util of `addOwnerAnnotation` does not capture if a change was or wasn't made. As such you might have added an annotation but didn't change anything about resources then the `isChanged` flag will be false and you will not persist your annotation change. That will be a bug

controllers/datafoundation.go
```diff
@@ -19,35 +30,58 @@ type dataFoundationSpec struct {
 type dataFoundationReconciler struct {
 	*ManagedFusionOfferingReconciler
 
+	offeringName                  string
+	offeringNamespace             string
```
nb-ohad, line 34: As we are under the context of a single offering, we might just want to use the offering name instead.
```suggestion
	dfName                  string
	dfNamespace             string
```

nb-ohad (changes requested):

controllers/managedfusionoffering_controller.go

controllers/managedfusionoffering_controller.go

controllers/datafoundation.go

controllers/datafoundation.go

controllers/datafoundation.go
```diff
@@ -195,3 +227,55 @@ func (r *dataFoundationReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDev
...
+	if err := r.list(&csvList); err != nil {
+		return fmt.Errorf("unable to list csv resources: %v", err)
+	}
+
+	for index := range csvList.Items {
+		csv := &csvList.Items[index]
+		if strings.HasPrefix(csv.Name, ocsOperatorName) {
+			if err := r.updateOCSCSV(csv); err != nil {
+				return fmt.Errorf("failed to update OCS CSV: %v", err)
+			}
```
nb-ohad, line 243: Why are we calling an update operation inside the loop? use the loop to find the CSV and then act on it after

controllers/datafoundation.go
```diff
@@ -195,3 +227,55 @@ func (r *dataFoundationReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDev
...
+			break
+		}
+	}
```
nb-ohad, line 246: I think we have a util to find a CR by prefix, can't we just use that?

controllers/managedfusionoffering_controller.go
```diff
@@ -133,7 +135,7 @@ func (r *ManagedFusionOfferingReconciler) reconcileOperatorGroup() error {
 		if err := r.own(r.operatorGroup); err != nil {
 			return err
 		}
-		r.operatorGroup.Spec = pluginGetDesiredOperatorGroupSpec(r)
+		r.operatorGroup.Spec = pluginGetDesiredOperatorGroupSpec(r.managedFusionOffering)
```
nb-ohad, line 138: Not sure why you decided to stop sending the reconciler, please add it back
```suggestion
		r.operatorGroup.Spec = pluginGetDesiredOperatorGroupSpec(r, r.managedFusionOffering)
```

controllers/managedfusionoffering_controller.go
```diff
@@ -149,7 +151,7 @@ func (r *ManagedFusionOfferingReconciler) reconcileCatalogSource() error {
 		if err := r.own(r.catalogSource); err != nil {
 			return err
 		}
-		desiredCatalogSourceSpec := pluginGetDesiredCatalogSourceSpec(r)
+		desiredCatalogSourceSpec := pluginGetDesiredCatalogSourceSpec(r.managedFusionOffering)
```
nb-ohad, line 154: Not sure why you decided to stop sending the reconciler, please add it back
```suggestion
		desiredCatalogSourceSpec := pluginGetDesiredCatalogSourceSpec(r, r.managedFusionOffering)
```

controllers/managedfusionoffering_controller.go
```diff
@@ -167,7 +169,7 @@ func (r *ManagedFusionOfferingReconciler) reconcileSubscription() error {
 		if err := r.own(r.subscription); err != nil {
 			return err
 		}
-		desiredSubscriptionSpec := pluginGetDesiredSubscriptionSpec(r)
+		desiredSubscriptionSpec := pluginGetDesiredSubscriptionSpec(r.managedFusionOffering)
```
nb-ohad, line 172: Not sure why you decided to stop sending the reconciler, please add it back
```suggestion
		desiredSubscriptionSpec := pluginGetDesiredSubscriptionSpec(r, r.managedFusionOffering)
```

nb-ohad (changes requested):

controllers/datafoundation.go
```diff
@@ -195,3 +227,65 @@ func (r *dataFoundationReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDev
...
+				}
+			case "ocs-metrics-exporter":
+				resources := datafoundation.GetResourceRequirements("ocs-metrics-exporter")
+				if !equality.Semantic.DeepEqual(container.Resources, resources) {
+					container.Resources = resources
+					isChanged = true
+				}
+			default:
+				r.Log.V(-1).Info("Could not find resource requirement", "Resource", container.Name)
+			}
```
nb-ohad, line 282: There is a repeated pattern here which obviously holds true for at least 3 cases, so I think we should collapse these 3 cases.
```suggestion
			switch container := &containers[j]; container.Name {
			case "ocs-operator",  "rook-ceph-operator",  "ocs-metrics-exporter":
				resources := datafoundation.GetResourceRequirements(container.Name)
				if !equality.Semantic.DeepEqual(container.Resources, resources) {
					container.Resources = resources
					isChanged = true
				}
			default:
				r.Log.V(-1).Info("Could not find resource requirement", "Resource", container.Name)
			}
```

And if the current assumption is that all containers will follow the same pattern then the preferred way is an `if` statement

```suggestion
			container := &containers[j]; container.Name 
			if container.Name == "ocs-operator" || container.Name == "rook-ceph-operator" || container.Name == "ocs-metrics-exporter" {
				resources := datafoundation.GetResourceRequirements(container.Name)
				if !equality.Semantic.DeepEqual(container.Resources, resources) {
					container.Resources = resources
					isChanged = true
				}
			else {
				r.Log.V(-1).Info("Could not find resource requirement", "Resource", container.Name)
			}
```

But it seems that the source of truth that derives this check is the CSV, not a static list which means a better solution might be
```suggestion
			container := &containers[j]; container.Name 
			resources := datafoundation.GetResourceRequirements(container.Name)
			if !equality.Semantic.DeepEqual(container.Resources, resources) {
				container.Resources = resources
				isChanged = true		
			}
```

As the assumption was always that we should panic and exist if we don't have a resource definition for a container and all containers should have a resource definition. That was done by design so we would capture these kinds of issues during development and not runtime!

nb-ohad (changes requested):

controllers/datafoundation.go
bindrad: Okay, updated it

controllers/managedfusionoffering_controller.go
```diff
@@ -180,11 +182,33 @@ func (r *ManagedFusionOfferingReconciler) reconcileSubscription() error {
...
+	if value, found := annotations[managedFusionAnnotation]; !found ||
+		value != fmt.Sprintf("%s/%s", r.namespace, r.managedFusionOffering.Name) {
+		utils.AddAnnotation(
+			res,
+			managedFusionAnnotation,
+			fmt.Sprintf("%s/%s", r.namespace, r.managedFusionOffering.Name),
+		)
+		return true
+	}
+	return false
```
nb-ohad, line 196: The logic here will be valuable for every annotation addition. Can we move it into the `AddAnnotation` util?
Also, I think this should become a general util function instead of a reconciler function

```
func addOwnerAnnotation(res client.Object, owner client.Object) bool {
	annotations := res.GetAnnotations()
	desiredAnnotationValue := fmt.Sprintf("%s/%s", owner.GetNamespace(), owner.GetName())
	return addAnnotation(res, ownerAnnotation, desiredAnnotationValue)
}
```

controllers/datafoundation.go
```diff
@@ -214,6 +244,49 @@ func (r *dataFoundationReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDev
...
+	deployments := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs
+	for i := range deployments {
+		containers := deployments[i].Spec.Template.Spec.Containers
+		for j := range containers {
+			container := &containers[j]
+			resources := datafoundation.GetResourceRequirements(container.Name)
+			if !equality.Semantic.DeepEqual(container.Resources, resources) {
+				container.Resources = resources
+				isChanged = true
+			}
```
nb-ohad, line 279: What happened to the warning that was here?

nb-ohad (changes requested):

controllers/managedfusionoffering_controller.go
bindrad: Moving logic to AddAnnotation won't be good because we have reconcile functions where the object gets updated under CreateOrUpdate function of controller runtime where this logic is already taken care of.

bindrad: But in this case, we will need to pass the owner object which will be managedFusionOffering CR and we are assuming that the offering's reconciler won't have access to it. So I think whatever it is right now is good.
Or we can have a function that will look something like this in utils package
```
func AddOwnerAnnotation(obj metav1.Object, ownerNamespace string, ownerName string, key string) bool {
	annotations := obj.GetAnnotations()
	desiredAnnotationValue := fmt.Sprintf("%s/%s", ownerNamespace, ownerName)
	if value, found := annotations[key]; !found ||
		value != desiredAnnotationValue {
		AddAnnotation(
			obj,
			key,
			desiredAnnotationValue,
		)
		return true
	}
	return false
}
```

controllers/datafoundation.go
```diff
@@ -214,6 +244,49 @@ func (r *dataFoundationReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDev
...
+		obj := &csvList.Items[index]
+		if strings.HasPrefix(obj.Name, ocsOperatorName) {
+			csv = obj
+			break
+		}
+	}
+	if csv == nil {
+		r.Log.Info("unable to find OCS CSV")
+		return nil
+	}
```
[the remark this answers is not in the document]

controllers/datafoundation.go
```diff
@@ -214,6 +244,49 @@ func (r *dataFoundationReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDev
...
+
+	isChanged := r.addOwnerAnnotation(csv)
+	deployments := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs
```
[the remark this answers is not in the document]

controllers/datafoundation.go
bindrad: Updated

utils/utils.go
```diff
@@ -58,13 +61,51 @@ func AddLabel(obj metav1.Object, key string, value string) {
...
 	if annotations == nil {
 		annotations = make(map[string]string)
 		obj.SetAnnotations(annotations)
+	} else if curr, found := annotations[key]; !found || curr != value {
+		annotations[key] = value
+		return true
 	}
-	annotations[key] = value
+	return false
+}
```
nb-ohad, line 74: There is a bug here. If there is no annotation map you are not adding it you are just creating the annotation map. Also, there is no reason to use `make` here  

```suggestion
func AddAnnotation(obj metav1.Object, key string, value string) bool {
	annotations := obj.GetAnnotations()
	if annotations == nil {
		annotations = {map[string]string}
		obj.SetAnnotations(annotations)
	} 
	if curr, found := annotations[key]; !found || curr != value {
		annotations[key] = value
		return true
	}
	return false
}
```

utils/utils.go
```diff
@@ -58,13 +61,51 @@ func AddLabel(obj metav1.Object, key string, value string) {
...
+			ref.Kind == owners[i].Kind &&
+			ref.Name == owners[i].Name &&
+			ref.UID == owners[i].UID {
+			index = i
+			break
+		}
+	}
+	if index != -1 {
+		owners[index] = ref
+		isChanged = false
```
nb-ohad, line 102: Why is this set to false? you are editing the annotation. 
This entire function is either appending an annotation or changing an existing one. 
Why do we want to track a change in the first place? I couldn't find a single call site that need this info and in fact all of the call sites are ignoring it using `_`

nb-ohad (changes requested):

utils/utils.go
bindrad: We are using it over here
https://github.com/red-hat-storage/managed-fusion-agent/pull/20/files#diff-8755ab6909c6e17d04613ea6d890a985c19a2ac23db5893a220fb226918782d5R248

controllers/datafoundation.go
```diff
@@ -214,9 +236,42 @@ func (r *dataFoundationReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDev
...
+		for i := range deployments {
+			containers := deployments[i].Spec.Template.Spec.Containers
+			for j := range containers {
+				container := &containers[j]
+				resources := datafoundation.GetResourceRequirements(container.Name)
+				if !equality.Semantic.DeepEqual(container.Resources, resources) {
+					container.Resources = resources
+				} else {
+					r.Log.V(-1).Info("Could not find resource requirement", "Resource", container.Name)
+				}
```
nb-ohad, line 261: The `else` content doesn't match the `if` content.  Also, why do we need a check for  `DeepEqual` if `GetAndUpdate` already takes care of it?

controllers/managedfusionoffering_controller.go
```diff
@@ -129,8 +134,8 @@ func (r *ManagedFusionOfferingReconciler) reconcilePhases() (reconcile.Result, e
 func (r *ManagedFusionOfferingReconciler) reconcileOperatorGroup() error {
 	r.Log.Info(fmt.Sprintf("Reconciling operator group for %s offering deployment", r.managedFusionOffering.Spec.Kind))
 
-	_, err := ctrl.CreateOrUpdate(r.Ctx, r.Client, r.operatorGroup, func() error {
-		if err := r.own(r.operatorGroup); err != nil {
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.operatorGroup, func() error {
+		if err := r.own(r.operatorGroup, true); err != nil {
```
nb-ohad, line 138:
```suggestion
	_, err := r.CreateOrUpdate(r.operatorGroup, func() error {
		if err := r.own(r.operatorGroup, true); err != nil {
```

controllers/managedfusionoffering_controller.go
```diff
@@ -145,8 +150,8 @@ func (r *ManagedFusionOfferingReconciler) reconcileOperatorGroup() error {
 func (r *ManagedFusionOfferingReconciler) reconcileCatalogSource() error {
 	r.Log.Info(fmt.Sprintf("Reconciling catalog source for %s offering deployment", r.managedFusionOffering.Spec.Kind))
 
-	_, err := ctrl.CreateOrUpdate(r.Ctx, r.Client, r.catalogSource, func() error {
-		if err := r.own(r.catalogSource); err != nil {
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.catalogSource, func() error {
+		if err := r.own(r.catalogSource, true); err != nil {
```
nb-ohad, line 154:
```suggestion
	_, err := r.CreateOrUpdate(r.catalogSource, func() error {
		if err := r.own(r.catalogSource, true); err != nil {
```

controllers/managedfusionoffering_controller.go
```diff
@@ -163,8 +168,8 @@ func (r *ManagedFusionOfferingReconciler) reconcileCatalogSource() error {
 func (r *ManagedFusionOfferingReconciler) reconcileSubscription() error {
 	r.Log.Info(fmt.Sprintf("Reconciling subscription for %s offering deployment", r.managedFusionOffering.Spec.Kind))
 
-	_, err := ctrl.CreateOrUpdate(r.Ctx, r.Client, r.subscription, func() error {
-		if err := r.own(r.subscription); err != nil {
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.subscription, func() error {
```
nb-ohad, line 171:
```suggestion
	_, err := r.CreateOrUpdate(r.subscription, func() error {
```

controllers/managedfusionoffering_controller.go
```diff
@@ -180,14 +185,65 @@ func (r *ManagedFusionOfferingReconciler) reconcileSubscription() error {
...
+	}
+	if err := r.update(obj); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *ManagedFusionOfferingReconciler) list(obj client.ObjectList, listOptions ...client.ListOption) error {
+	return r.Client.List(r.ctx, obj, listOptions...)
+}
```
nb-ohad, line 219: Can we please move these 2 after the simple operations (get, list, update, etc.)?

nb-ohad (changes requested):

nb-ohad (dismissed):

nb-ohad (approved):

PR #22, opened 2023-03-20
Use typed assignment instead of declaring floating variables

bindrad:


controllers/managedfusionoffering_controller.go
```diff
@@ -213,7 +213,8 @@ func parseDataFoundationSpec(r *ManagedFusionOfferingReconciler) (dataFoundation
 	valid := true
```
nb-ohad, line 213: Lets change that to `isValid`

controllers/managedfusionoffering_controller.go
```diff
@@ -213,7 +213,8 @@ func parseDataFoundationSpec(r *ManagedFusionOfferingReconciler) (dataFoundation
...
 	var usableCapacityInTiB int
 	var err error
```
nb-ohad, line 215: Reverse the order.
```suggestion
	var err error
	var usableCapacityInTiB int
```

nb-ohad (changes requested):

leelavg (approved):

ezio-auditore (approved):

nb-ohad (approved):

PR #23, opened 2023-03-23
Add reconcile function for OCSInitialization 

bindrad:
This PR will enable running of ceph toolbox pod for DF offering

### DIFF
--- a/controllers/datafoundation.go
+++ b/controllers/datafoundation.go
@@ -41,11 +41,13 @@ type dataFoundationReconciler struct {
 	cephIngressNetworkPolicy       *netv1.NetworkPolicy
 	providerAPIServerNetworkPolicy *netv1.NetworkPolicy
 	rookConfigMap                  *corev1.ConfigMap
+	ocsInitialization              *ocsv1.OCSInitialization
 }
 
 //+kubebuilder:rbac:groups=ocs.openshift.io,namespace=system,resources=storageclusters,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=network.openshift.io,resources=ingressnetworkpolicies,verbs=create;get;list;watch;update
 //+kubebuilder:rbac:groups=operators.coreos.com,namespace=system,resources=clusterserviceversions,verbs=get;list;watch;delete;update;patch
+//+kubebuilder:rbac:groups=ocs.openshift.io,namespace=system,resources=ocsinitializations,verbs=get;list;watch;update;patch
 
 func dfSetupWatches(controllerBuilder *builder.Builder) {
 	csvPredicates := builder.WithPredicates(
@@ -60,7 +62,8 @@ func dfSetupWatches(controllerBuilder *builder.Builder) {
 		Owns(&ocsv1.StorageCluster{}).
 		Owns(&netv1.NetworkPolicy{}).
 		Owns(&corev1.ConfigMap{}).
-		Owns(&opv1a1.ClusterServiceVersion{}, csvPredicates)
+		Owns(&opv1a1.ClusterServiceVersion{}, csvPredicates).
+		Owns(&ocsv1.OCSInitialization{})
 }
 
 func DFAddToScheme(scheme *runtime.Scheme) {
@@ -90,6 +93,9 @@ func dfReconcile(offeringReconciler *ManagedFusionOfferingReconciler, offering *
 		return err
 	}
 	if err := r.reconcileRookCephOperatorConfig(); err != nil {
+		return err
+	}
+	if err := r.reconcileOCSInitialization(); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


controllers/datafoundation.go
```diff
@@ -316,3 +324,32 @@ func (r *dataFoundationReconciler) reconcileProviderAPIServerNetworkPolicy() err
...
+	if !obj.Spec.EnableCephTools {
+		isChanged = true
+		obj.Spec.EnableCephTools = true
+	}
+	isChanged = r.addOwnerAnnotation(obj)
+	if isChanged {
+		if err := r.update(obj); err != nil {
+			return err
+		}
+	}
```
nb-ohad, line 352: We do not use owner annotation anymore, that feature suggestion was never merged into main. 
We decided to go with non-controller owner refs
```suggesion
	csv := &ocsInitList.Items[0]
	if err := r.Own(csv, false); err != nil {
		return err
	}
	if !csv.Spec.EnableCephTools {
		obj.Spec.EnableCephTools = csv
		if err := r.update(csv); err != nil {
			return err
		}
	}
```

nb-ohad (changes requested):

controllers/datafoundation.go
```diff
@@ -329,3 +334,29 @@ func (r *dataFoundationReconciler) reconcileRookCephOperatorConfig() error {
...
+
+	ocsInitList := ocsv1.OCSInitializationList{}
+	if err := r.list(&ocsInitList, client.InNamespace(r.dfNamespace)); err != nil {
+		return fmt.Errorf("failed to list OCSInitialization resources: %v", err)
+	}
+	if len(ocsInitList.Items) == 0 {
+		r.Log.V(-1).Info("OCSInitialization resource not found")
+		return nil
+	}
+	obj := &ocsInitList.Items[0]
```
nb-ohad, line 349: If we only need the first one, why aren't we limiting the list to a single result? This will make sure that k8s will only fetch one record from its internal DB

controllers/datafoundation.go
```diff
@@ -329,3 +334,29 @@ func (r *dataFoundationReconciler) reconcileRookCephOperatorConfig() error {
...
+	err := r.GetAndUpdate(obj, func() error {
```
nb-ohad, line 350: It does not make sense to do GetAndUpdate if we already fetched the resource. Why can't we just let GetAndUpdate fetch the resource for us using the name of the resource?

nb-ohad (changes requested):

controllers/datafoundation.go
leelavg, line 350: - why is there a `GetAndUpdate` function in the first place
- what info are we missing in the supplied `obj` that needs explicit `get` again?

controllers/datafoundation.go
```diff
@@ -329,3 +334,29 @@ func (r *dataFoundationReconciler) reconcileRookCephOperatorConfig() error {
 
 	return nil
 }
+
+func (r *dataFoundationReconciler) reconcileOCSInitialization() error {
+	r.Log.Info("Reconciling OCSInitialization")
+
+	ocsInitList := ocsv1.OCSInitializationList{}
```
leelavg, line 341: - will there be more than one CRs of this type?
- if not, will there be different names for these CRs across deployments?
- IOW, if there's a single CR of this CRD and static name across deployment, why can't just get it directly?

controllers/datafoundation.go

bindrad: > why is there a GetAndUpdate function in the first place
Because there are few reconcile functions where we don't want to update when the obj is in the desired state.
I have updated the code so I'm not explicitly getting the object again.

controllers/datafoundation.go
bindrad: > will there be more than one CRs of this type?

No, it will always be a single CR of this type. It is created when OCS-Operator is installed.
> if not, will there be different names for these CRs across deployments?

No, name of this CR will always be ocsinit
> IOW, if there's a single CR of this CRD and static name across deployments, why can't just get it directly?

Yes, updated it in the latest commit.

controllers/datafoundation.go
nb-ohad: @leelavg GetAndUpdate should be the recommended way for this point forward to get an object, mutate it using a mutate function, assert changes (if any), and update the object (if needed). 

This is the result of us observing almost the same pattern repeating itself in multipole reconcile functions.
And from code structure perspective we followed the `CreateAndUpdate` pattern (without the create part)

nb-ohad (approved):

controllers/datafoundation.go
leelavg: > there are few reconcile functions where we don't want to update when the obj is in the desired state.
- ack, I thought sdk already checks the equality before update

nb-ohad (approved):

Issue #28, opened 2023-03-24
ManagedFusionOffering Controller expects Offering CRDs to be present on the cluster to setup watches


bindrad:
ManagedFusionOffering Controller expects storageCluster CRD to be present on the cluster to sync the cache. As we are using a single controller to install and configure the offering, we need to watch the objects that should be configured as a part of offering installation(for example: storageCluster). StorageCluster CRD is not even present on the cluster as OCS Operator is not installed yet.
This is the error in Agent pod
` 2023-03-24T09:53:27.789Z ERROR setup Problem running manager {"error": "failed to wait for managedfusionoffering caches to sync: no matches for kind \"StorageCluster\" in version \"ocs.openshift.io/v1\""} `

PR #29, opened 2023-03-24
Exit when CRDs are found for the first time to register with watches

bindrad:
Problem description can be found [here](https://github.com/red-hat-storage/managed-fusion-agent/issues/28)

To build the cache CRDs for the offering should exist in the cluster but Agent installs the offering and then CRDs are created into the cluster. Because of this, the manager was not able to start.

So as a quick fix, we can run the controller initially to just install the offering(without watching offering resources) and then kill the container and re-run it with the offering watches.

### DIFF
--- a/controllers/managedfusionoffering_controller.go
+++ b/controllers/managedfusionoffering_controller.go
@@ -19,16 +19,22 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"os"
 
 	"github.com/go-logr/logr"
 	opv1 "github.com/operator-framework/api/pkg/operators/v1"
 	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	v1alpha1 "github.com/red-hat-storage/managed-fusion-agent/api/v1alpha1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	controller "sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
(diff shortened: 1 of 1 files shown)


nb-ohad:
Please add to the description a link to the github issue capturing the problem

controllers/managedfusionoffering_controller.go
```diff
@@ -65,11 +72,42 @@ func (r *ManagedFusionOfferingReconciler) SetupWithManager(mgr ctrl.Manager, ctr
 		Owns(&opv1a1.CatalogSource{}).
 		Owns(&opv1a1.Subscription{})
 
-	pluginSetupWatches(controllerBuilder)
+	if doesCRDsExists() {
+		r.DoesCRDExistsAtStart = true
+		pluginSetupWatches(controllerBuilder)
+	}
```
nb-ohad, line 78: We have an established way to check for CRD existence. It was added with the SDN/OVN contribution a couple of months back on the original deployer repo. Please use this

https://github.com/red-hat-storage/ocs-osd-deployer/blob/main/controllers/managedocs_controller.go#L323-L328


Also, why is this implemented at the offering reconciler level where it is a fix for df add watches? This should go into dfSetupWatches. This all hack should be internal to the DF offering.

controllers/managedfusionoffering_controller.go
```diff
@@ -65,11 +72,42 @@ func (r *ManagedFusionOfferingReconciler) SetupWithManager(mgr ctrl.Manager, ctr
...
+		if errors.IsNotFound(err) {
+			return false
+		} else {
+			setupLog.Error(err, fmt.Sprintf("Error Getting CRD for %s", "storageclusters.ocs.openshift.io"))
+			os.Exit(1)
+		}
+	}
+
+	return true
+}
```
nb-ohad, line 109: As of the last comment
```suggestion
```

controllers/managedfusionoffering_controller.go
```diff
@@ -84,11 +122,20 @@ func (r *ManagedFusionOfferingReconciler) Reconcile(ctx context.Context, req ctr
...
 	} else {
-		return result, nil
+		if result, err := pluginReconcile(r, r.managedFusionOffering); err != nil {
+			return ctrl.Result{}, fmt.Errorf("An error was encountered during reconcile: %v", err)
+		} else {
+			return result, nil
+		}
 	}
+
+	return ctrl.Result{}, nil
```
nb-ohad, line 138: All of this code is very specific for DF, even if it does not say DF in the variable names, we should not do that. For this workaround, only the actual offering should know if it needs a restart or not and that code should go into the dfReconcile code at the start. But the reconciler should implement an `r.Exist()` function to allow the offering to exit without using a hack.  
```suggestion
	if result, err := pluginReconcile(r, r.managedFusionOffering); err != nil {
		return ctrl.Result{}, fmt.Errorf("An error was encountered during reconcile: %v", err)
	} else {
		return result, nil
	}
```

nb-ohad (changes requested):

controllers/managedfusionoffering_controller.go
```diff
@@ -51,6 +57,7 @@ type ManagedFusionOfferingReconciler struct {
 	operatorGroup         *opv1.OperatorGroup
 	catalogSource         *opv1a1.CatalogSource
 	subscription          *opv1a1.Subscription
+	DoesCRDExistsAtStart  bool
```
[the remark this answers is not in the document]
nb-ohad: @leelavg You cannot wait in an init container for something that the same entity needs to initiate the installation for

bindrad:
https://github.com/red-hat-storage/managed-fusion-agent/pull/35 implements the workaround so closing this PR

leelavg:
should be closed in favour of #54?

bindrad:
closed in favour of https://github.com/red-hat-storage/managed-fusion-agent/pull/54

PR #30, opened 2023-03-27
Added DataFoundation Client Offering

ezio-auditore:
Added DataFoundation Client offering with the following reconcile funcs and dependent go packages

-  reconcileStorageClient
-  reconcileCSV
- reconcileDefaultBlockStorageClassClaim
- reconcileDefaultFSStorageClassClaim


controllers/datafoundation_client.go
```diff
@@ -0,0 +1,134 @@
...
+func (r *dataFoundationClientReconciler) initReconciler(reconciler *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) {
+	r.ManagedFusionOfferingReconciler = reconciler
+	r.offering = offering
+}
+
+func (r *dataFoundationClientReconciler) parseSpec(offering *v1alpha1.ManagedFusionOffering) error {
+	r.Log.Info("Parsing ManagedFusionOffering Data Foundation Client spec")
+
+	isValid := true
+	onboardingTicketAsString, found := offering.Spec.Config["onboarding_ticket"]
```
nb-ohad, line 60: Onboarding ticket is a string no need to mention as string
```suggestion
	onboardingTicket, found := offering.Spec.Config["onboarding_ticket"]
```

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,134 @@
...
+		isValid = false
+	} else if onboardingTicketAsString == "" {
+		r.Log.Error(
+			fmt.Errorf("onboarding_ticket is empty "),
+			"an error occurred while parsing ManagedFusionOffering Data Foundation Client spec",
+		)
+		isValid = false
+	}
+
+	storageDFEndpointAsString, found := offering.Spec.Config["storage_DF_Endpoint"]
```
nb-ohad, line 75: storageDFEndpoint is a string unless you want to parse it as an IP. 
Also the key name and variable do not make sense
```suggestion
	dfProviderEndpoint, found := offering.Spec.Config["data_foundation_provider_endpoint"]
```

datafoundationclient/templates_storageclient.go
nb-ohad, line 1: Why do we need an empty template? if all the information is coming from the outside there is no need for a template at all

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,134 @@
...
+		if err := r.reconcileStorageClient(r.offering); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *dataFoundationClientReconciler) reconcileStorageClient(offering *v1alpha1.ManagedFusionOffering) error {
+
```
nb-ohad, line 115:
```suggestion
```

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,134 @@
...
+type dataFoundationClientSpec struct {
+	onboardingTicket  string
+	storageDFEndpoint string
+}
+
+type dataFoundationClientReconciler struct {
+	*ManagedFusionOfferingReconciler
+
+	offering      *v1alpha1.ManagedFusionOffering
+	spec          dataFoundationClientSpec
```
nb-ohad, line 25: Is this a good name for a variable in the reconciler?

nb-ohad (changes requested):

controllers/datafoundation_client.go
ezio-auditore: updated

controllers/datafoundation_client.go
ezio-auditore: updated

datafoundationclient/templates_storageclient.go
ezio-auditore: updated

controllers/datafoundation_client.go
ezio-auditore: updated

controllers/datafoundation_client.go
ezio-auditore: updated

main.go
```diff
@@ -195,4 +195,5 @@ func mapCRDAvailability(crdNames ...string) map[string]bool {
 // This function is a placeholder for offering plugin integration
 func pluginAddToScheme(scheme *runtime.Scheme) {
 	controllers.DFAddToScheme(scheme)
+	controllers.DFCAddToScheme(scheme)
```
leelavg, line 198: - maybe a TODO comment here will be good stating atm we can't run the offering reconciler standalone?
- similar one at setupwatches as well

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,170 @@
...
+
+	r.storageclient.Name = "ocs-storageclient"
+	r.storageclient.Namespace = offering.Namespace
+}
+
+func (r *dataFoundationClientReconciler) parseSpec(offering *v1alpha1.ManagedFusionOffering) error {
+	r.Log.Info("Parsing ManagedFusionOffering Data Foundation Client spec")
+
+	isValid := true
+	onboardingTicket, found := offering.Spec.Config["onboarding_ticket"]
```
bindrad, line 79:
```suggestion
	onboardingTicket, found := offering.Spec.Config["onboardingTicket"]
```
Use camelcase for the keys in the config of offering CR so that we can maintain consistency across all offerings

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,170 @@
...
+	if !found {
+		r.Log.Error(
+			fmt.Errorf("missing field: onboarding_ticket"),
```
bindrad, line 82:
```suggestion
			fmt.Errorf("missing field: onboardingTicket"),
```

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,170 @@
...
+			"an error occurred while parsing ManagedFusionOffering Data Foundation Client spec",
+		)
+		isValid = false
+	} else if onboardingTicket == "" {
+		r.Log.Error(
+			fmt.Errorf("onboarding_ticket is empty "),
```
bindrad, line 88:
```suggestion
			fmt.Errorf("onboardingTicket is empty"),
```

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,170 @@
...
+			"an error occurred while parsing ManagedFusionOffering Data Foundation Client spec",
+		)
+		isValid = false
+	}
+
+	dfstorageProviderEndpoint, found := offering.Spec.Config["data_foundation_provider_endpoint"]
```
bindrad, line 94:
```suggestion
	dfStorageProviderEndpoint, found := offering.Spec.Config["endpoint"]
```
I think the `endpoint` is enough because the endpoint is a specific detail related to DFC offering only. If you think we should have complete `data_foundation_provider_endpoint` then please make it camel-cased

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,170 @@
...
+	if !found {
+		r.Log.Error(
+			fmt.Errorf("missing field: data_foundation_provider_endpoint"),
```
bindrad, line 97: Same as above

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,170 @@
...
+			"an error occurred while parsing ManagedFusionOffering Data Foundation Client spec",
+		)
+		isValid = false
+	} else if dfstorageProviderEndpoint == "" {
+		r.Log.Error(
+			fmt.Errorf("data_foundation_provider_endpoint is empty "),
```
bindrad, line 103: Same as above

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,170 @@
...
+		dfstorageProviderEndpoint: dfstorageProviderEndpoint,
+	}
+	return nil
+}
+
+func (r *dataFoundationClientReconciler) reconcilePhases() (ctrl.Result, error) {
+	if !r.offering.DeletionTimestamp.IsZero() {
+		return ctrl.Result{}, nil
+	} else {
+		if err := r.reconcileStorageClient(r.offering); err != nil {
```
bindrad, line 126:
```suggestion
		if err := r.reconcileStorageClient(); err != nil {
```
Why are you passing offering to reconcileStorageClient? You can access it from `dataFoundationClientReconciler`

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,170 @@
...
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+}
+
+func (r *dataFoundationClientReconciler) reconcileStorageClient(offering *v1alpha1.ManagedFusionOffering) error {
+
+	r.Log.Info("Reconciling StorageClient")
+	_, err := r.CreateOrUpdate(&r.storageclient, func() error {
```
bindrad, line 139:
```suggestion
func (r *dataFoundationClientReconciler) reconcileStorageClient(offering *v1alpha1.ManagedFusionOffering) error {
	r.Log.Info("Reconciling StorageClient")

	_, err := r.CreateOrUpdate(&r.storageclient, func() error {
```

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,170 @@
...
+		if err := r.own(&r.storageclient, true); err != nil {
```
bindrad, line 140:
```suggestion
		if err := r.own(&r.storageClient, true); err != nil {
```

bindrad: Same change wherever we are using `storageclient` instead of `storageClient`

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,170 @@
...
+
+	if !isValid {
+		r.Log.Info("parsing ManagedFusionOffering Data Foundation  Client spec failed")
+		return fmt.Errorf("invalid ManagedFusionOffering Data Foundation Client spec")
+	}
+	r.Log.Info("parsing ManagedFusionOffering Data Foundation Client spec completed successfuly")
+
+	r.dataFoundationClientSpec = dataFoundationClientSpec{
+		onboardingTicket:          onboardingTicket,
+		dfstorageProviderEndpoint: dfstorageProviderEndpoint,
```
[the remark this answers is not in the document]
bindrad: Same change wherever we are using `dfstorageProviderEndpoint` instead of `dfStorageProviderEndpoint`

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,170 @@
...
+func (r *dataFoundationClientReconciler) reconcileCSV() error {
+	r.Log.Info("Reconciling OCS Client Operator CSV")
+
+	csv, err := r.getCSVByPrefix(ocsClientOperatorName)
+	if err != nil {
+		return fmt.Errorf("unable to find OCS client CSV: %v", err)
+	}
+	err = r.GetAndUpdate(csv, func() error {
+		if err := r.own(csv, false); err != nil {
+			return fmt.Errorf("unable to set ownerRef on ocs  client csv: %v", err)
```
bindrad, line 162:
```suggestion
			return fmt.Errorf("unable to set ownerRef on ocs client csv: %v", err)
```

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,170 @@
...
+	r.Log.Info("parsing ManagedFusionOffering Data Foundation Client spec completed successfuly")
+
+	r.dataFoundationClientSpec = dataFoundationClientSpec{
+		onboardingTicket:          onboardingTicket,
+		dfstorageProviderEndpoint: dfstorageProviderEndpoint,
+	}
+	return nil
+}
+
+func (r *dataFoundationClientReconciler) reconcilePhases() (ctrl.Result, error) {
```
bindrad, line 122: How are we going to create default storageClassClaims?

bindrad, line 122: I think the uninstallation code is not implemented yet, right?

controllers/managedfusionoffering_controller.go
```diff
@@ -305,8 +316,19 @@ func pluginGetDesiredCatalogSourceSpec(r *ManagedFusionOfferingReconciler) opv1a
...
+			Package: "ocs-operator",
+		}
+	case v1alpha1.KindDataFoundationClient:
+		desiredSubscription = &opv1a1.SubscriptionSpec{
+			Channel: "stable-4.12",
+			Package: "ocs-client-operator",
+		}
 	}
+
+	return desiredSubscription
```
bindrad, line 333:
```suggestion
	switch r.managedFusionOffering.Spec.Kind {
	case v1alpha1.KindDataFoundation:
		return &opv1a1.SubscriptionSpec{
			Channel: "stable-4.12",
			Package: "ocs-operator",
		}
	case v1alpha1.KindDataFoundationClient:
		return &opv1a1.SubscriptionSpec{
			Channel: "stable-4.12",
			Package: "ocs-client-operator",
		}
	}
```

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,170 @@
...
+			"an error occurred while parsing ManagedFusionOffering Data Foundation Client spec",
+		)
+		isValid = false
+	} else if dfstorageProviderEndpoint == "" {
+		r.Log.Error(
+			fmt.Errorf("data_foundation_provider_endpoint is empty "),
+			"an error occurred while parsing ManagedFusionOffering Data Foundation Client spec",
+		)
+		isValid = false
+	}
```
bindrad, line 107: Shouldn't we validate `endpoint`?
You can use the same method we use to validate smtp endpoint in managed fusion controller

bindrad (changes requested):

controllers/datafoundation_client.go
ezio-auditore: Updated

controllers/datafoundation_client.go
ezio-auditore: I am using `dataFoundationProviderEndpoint` for the key instead of `endpoint`

controllers/datafoundation_client.go
ezio-auditore: Updated

controllers/datafoundation_client.go
ezio-auditore: Yes correct. Updated

controllers/datafoundation_client.go
ezio-auditore: Updated

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,170 @@
...
+	onboardingTicket          string
+	dfstorageProviderEndpoint string
+}
+
+type dataFoundationClientReconciler struct {
+	*ManagedFusionOfferingReconciler
+
+	offering                 *v1alpha1.ManagedFusionOffering
+	dataFoundationClientSpec dataFoundationClientSpec
+	storageclient            ocsclient.StorageClient
```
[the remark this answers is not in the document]
ezio-auditore: Updated

controllers/datafoundation_client.go
ezio-auditore: Updated

controllers/datafoundation_client.go
ezio-auditore: Updated

controllers/datafoundation_client.go
ezio-auditore: Uninstall flow not yet implemented in this PR

controllers/managedfusionoffering_controller.go
ezio-auditore: I am keeping it as it is, since the return statement will have a nil object which is not being handled in the managedFusionoffering_controller yet

controllers/datafoundation_client.go
ezio-auditore: Updated

controllers/datafoundation_client.go
ezio-auditore: Added storageclassclaims reconcile func

main.go
ezio-auditore: We already have a comment above stating the function is a `placeholder`.

Similar to setupWatches

controllers/datafoundation_client.go
bindrad: The error is still `missing field: onboarding_ticket` instead of `missing field: onboardingTicket`

bindrad (changes requested):

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,215 @@
...
+			Name:      r.storageClient.Name,
+			Namespace: r.storageClient.Namespace,
+		}
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("unable to create/update Default StorageClassClaim: %v", err)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
ezio-auditore: Added Filesystem as well

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,215 @@
...
+	if dfstorageProviderEndpointAsURL.Host == "" {
+		r.Log.Error(
+			fmt.Errorf("cannot retrieve host from %s, might be a malformed or partial URL", dfstorageProviderEndpoint),
+			"an error occurred while parsing ManagedFusionOffering Data Foundation Client spec",
+		)
+		isValid = false
+	}
+
+	if !isValid {
+		r.Log.Info("parsing ManagedFusionOffering Data Foundation  Client spec failed")
```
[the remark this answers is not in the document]
ezio-auditore: updated

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,215 @@
...
+		)
+		isValid = false
+	} else if dfstorageProviderEndpoint == "" {
+		r.Log.Error(
+			fmt.Errorf("dataFoundationProviderEndpoint is empty "),
+			"an error occurred while parsing ManagedFusionOffering Data Foundation Client spec",
+		)
+		isValid = false
+	}
+	dfstorageProviderEndpointAsURL, err := url.ParseRequestURI(dfstorageProviderEndpoint)
```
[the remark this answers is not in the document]
ezio-auditore: updated

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	dfstorageProviderEndpoint, found := offering.Spec.Config["dataFoundationProviderEndpoint"]
+	if !found {
+		r.Log.Error(
+			fmt.Errorf("missing field: dataFoundationProviderEndpoint"),
+			"an error occurred while parsing ManagedFusionOffering Data Foundation Client spec",
+		)
+		isValid = false
+	} else if dfstorageProviderEndpoint == "" {
```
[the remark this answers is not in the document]
ezio-auditore: Updated

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,215 @@
...
+		isValid = false
+	} else if onboardingTicket == "" {
+		r.Log.Error(
+			fmt.Errorf("onboardingTicket is empty "),
+			"an error occurred while parsing ManagedFusionOffering Data Foundation Client spec",
+		)
+		isValid = false
+	}
+
+	dfstorageProviderEndpoint, found := offering.Spec.Config["dataFoundationProviderEndpoint"]
```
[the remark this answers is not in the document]
ezio-auditore: Updated

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,215 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+const (
+	ocsClientOperatorName = "ocs-client-operator"
+)
+
+type dataFoundationClientSpec struct {
+	onboardingTicket          string
+	dfstorageProviderEndpoint string
```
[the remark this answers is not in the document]
ezio-auditore: updated

api/v1alpha1/groupversion_info.go
```diff
@@ -15,8 +15,8 @@ limitations under the License.
 */
 
 // Package v1alpha1 contains API Schema definitions for the misf v1alpha1 API group
-//+kubebuilder:object:generate=true
-//+groupName=misf.ibm.com
+// +kubebuilder:object:generate=true
+// +groupName=misf.ibm.com
```
bindrad, line 19: Why do we need this change?

controllers/datafoundation_client.go
bindrad: This is not fixed

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,233 @@
...
+func dfcSetupWatches(controllerBuilder *builder.Builder) {
+	csvPredicates := builder.WithPredicates(
+		predicate.NewPredicateFuncs(
+			func(res client.Object) bool {
+				return strings.HasPrefix(res.GetName(), ocsClientOperatorName)
+			},
+		),
+	)
+	controllerBuilder.
+		Owns(&corev1.Secret{}).
```
bindrad, line 52: Why do we need to watch secret?

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,233 @@
...
+		r.storageClient.Spec.StorageProviderEndpoint = r.dataFoundationClientSpec.dfStorageProviderEndpoint
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("unable to create/update StorageClient: %v", err)
+	}
+	return nil
+}
+
+func (r *dataFoundationClientReconciler) reconcileCSV() error {
```
bindrad, line 182: Why do we need this function?

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,233 @@
...
+		}
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to update OCS client CSV: %v", err)
+	}
+	return nil
+}
+
+func (r *dataFoundationClientReconciler) reconcileDefaultStorageClassClaim() error {
```
bindrad, line 201: The structure of code up until now is one resource reconciliation per function. So can you split this into two different functions?

bindrad (changes requested):

controllers/datafoundation_client.go
ezio-auditore: Removed it. We don't need to watch secret

api/v1alpha1/groupversion_info.go
ezio-auditore: wasn't this decided to not keep any whitespace between the comment slashes?

controllers/datafoundation_client.go
ezio-auditore: To set a ownerRef on the `ocs-client-operator` csv

controllers/datafoundation_client.go
ezio-auditore: Updated.Splitted into two funcs

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,215 @@
...
+
+func (r *dataFoundationClientReconciler) reconcileStorageClient() error {
+	r.Log.Info("Reconciling StorageClient")
+
+	_, err := r.CreateOrUpdate(&r.storageClient, func() error {
+		if err := r.own(&r.storageClient, true); err != nil {
+			return err
+		}
+		r.storageClient.Spec.OnboardingTicket = r.dataFoundationClientSpec.onboardingTicket
+		r.storageClient.Spec.StorageProviderEndpoint = r.dataFoundationClientSpec.dfstorageProviderEndpoint
```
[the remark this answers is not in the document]
ezio-auditore: I don't agree with the above naming since it is a Data Foundation provider endpoint, should be marked as such

api/v1alpha1/groupversion_info.go
bindrad: Why in this PR?
I checked other kubebuilder markers and we don't have a space in between.

controllers/datafoundation_client.go
bindrad: Why do we need to set ownerRef on client-operator csv? 
We are not reconciling anything in client-operator csv

bindrad (changes requested):

bindrad:
@ezio-auditore Don't we need to add resource limits on client operator resources?

api/v1alpha1/groupversion_info.go
ezio-auditore: Sure, I will remove that specific commit

ezio-auditore: Removed the lint commit

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,240 @@
...
+	})
+	if err != nil {
+		return fmt.Errorf("unable to create/update Default blockpool StorageClassClaim: %v", err)
+	}
+	return nil
+}
+
+func (r *dataFoundationClientReconciler) reconcileDefaultFSStorageClassClaim() error {
+	r.Log.Info("Reconciling Default Filesystem StorageClassClaim")
+	_, err := r.CreateOrUpdate(&r.defaultFSStorageClassClaim, func() error {
```
bindrad, line 224:
```suggestion
	r.Log.Info("Reconciling Default Filesystem StorageClassClaim")

	_, err := r.CreateOrUpdate(&r.defaultFSStorageClassClaim, func() error {
```

bindrad (changes requested):

controllers/datafoundation_client.go
ezio-auditore: updated

ezio-auditore:
> @ezio-auditore Don't we need to add resource limits on client operator resources?

Yes, but not in this PR. Also the the reconcile for adding ownerRef for the ocs-client-operator CSV is because of this.

controllers/datafoundation_client.go
bindrad: Resolving this because we will utilize the function to add resource limits to ocs-client operator resources.

bindrad (dismissed):

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,241 @@
...
+			"an error occurred while parsing ManagedFusionOffering Data Foundation Client spec",
+		)
+		isValid = false
+	} else if onboardingTicket == "" {
+		r.Log.Error(
+			fmt.Errorf("onboardingTicket is empty"),
+			"an error occurred while parsing ManagedFusionOffering Data Foundation Client spec",
+		)
+		isValid = false
+	}
```
nb-ohad, line 101: Why are we checking for values here? This is supposed to be the equivalent of parsing of unstructured data into structured one, which mean sementic checks only. The empty check will make more sense in the reconcile method that uses the onboarding ticket.

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,241 @@
...
+			"an error occurred while parsing ManagedFusionOffering Data Foundation Client spec",
+		)
+		isValid = false
+	} else if dfStorageProviderEndpoint == "" {
+		r.Log.Error(
+			fmt.Errorf("dataFoundationProviderEndpoint is empty"),
+			"an error occurred while parsing ManagedFusionOffering Data Foundation Client spec",
+		)
+		isValid = false
+	}
```
nb-ohad, line 116: Same as last comment

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,241 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+const (
+	ocsClientOperatorName = "ocs-client-operator"
+)
+
+type dataFoundationClientSpec struct {
+	onboardingTicket          string
+	dfStorageProviderEndpoint string
```
nb-ohad, line 25: No need for a df prefix, or even the word storage. This struct only exists only in the df client context.
```suggestion
	providerEndpoint string
```

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,241 @@
...
+		isValid = false
+	} else if onboardingTicket == "" {
+		r.Log.Error(
+			fmt.Errorf("onboardingTicket is empty"),
+			"an error occurred while parsing ManagedFusionOffering Data Foundation Client spec",
+		)
+		isValid = false
+	}
+
+	dfStorageProviderEndpoint, found := offering.Spec.Config["dataFoundationProviderEndpoint"]
```
nb-ohad, line 103: See one of the previous comments for explanation.  
```suggestion
	providerEndpoint, found := offering.Spec.Config["dataFoundationProviderEndpoint"]
```

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,241 @@
...
+		)
+		isValid = false
+	}
+	if dfStorageProviderEndpointAsURL.Host == "" {
+		r.Log.Error(
+			fmt.Errorf("cannot retrieve host from %s, might be a malformed or partial URL", dfStorageProviderEndpoint),
+			"an error occurred while parsing ManagedFusionOffering Data Foundation Client spec",
+		)
+		isValid = false
+	}
```
nb-ohad, line 131: This check is about the value itself, not the structure. I would suggest moving this check into the reconciliation method that uses the provider endpoint

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,241 @@
...
+
+	if !isValid {
+		r.Log.Info("parsing ManagedFusionOffering Data Foundation Client spec failed")
+		return fmt.Errorf("invalid ManagedFusionOffering Data Foundation Client spec")
+	}
+	r.Log.Info("parsing ManagedFusionOffering Data Foundation Client spec completed successfuly")
+
+	r.dataFoundationClientSpec = dataFoundationClientSpec{
+		onboardingTicket:          onboardingTicket,
+		dfStorageProviderEndpoint: dfStorageProviderEndpoint,
```
nb-ohad, line 141:
```suggestion
		providerEndpoint: providerEndpoint,
```

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+}
+
+func (r *dataFoundationClientReconciler) reconcileStorageClient() error {
+	r.Log.Info("Reconciling StorageClient")
+
+	_, err := r.CreateOrUpdate(&r.storageClient, func() error {
+		if err := r.own(&r.storageClient, true); err != nil {
+			return err
+		}
```
nb-ohad, line 172: Please add whatever check makes sense for the values before setting them on the spec.

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	return nil
+}
+
+func (r *dataFoundationClientReconciler) reconcileCSV() error {
+	r.Log.Info("Reconciling OCS Client Operator CSV")
+
+	csv, err := r.getCSVByPrefix(ocsClientOperatorName)
+	if err != nil {
+		return fmt.Errorf("unable to find OCS client CSV: %v", err)
```
nb-ohad, line 188:
```suggestion
		return fmt.Errorf("unable to get OCS client CSV: %v", err)
```

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	err = r.GetAndUpdate(csv, func() error {
+		if err := r.own(csv, false); err != nil {
+			return fmt.Errorf("unable to set ownerRef on ocs client csv: %v", err)
+		}
+		return nil
+	})
```
nb-ohad, line 195: Why do you use `getAndUpdate` if you already loaded what csv resource via `getCSVByPrefix`? 
You only need to update.

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	return nil
+}
+
+func (r *dataFoundationClientReconciler) reconcileDefaultBlockStorageClassClaim() error {
+	r.Log.Info("Reconciling Default BlockPool StorageClassClaim")
+
+	_, err := r.CreateOrUpdate(&r.defaultBlockStorageClassClaim, func() error {
+		if err := r.own(&r.defaultBlockStorageClassClaim, true); err != nil {
+			return fmt.Errorf("unable to set ownerRef on Default Blockpool StorageClassClaim: %v", err)
```
nb-ohad, line 207: No need for a custom error here as `r.own` will provide an error with a context. This is the same practice we follow everywhere else we use `r.own`.
```suggestion
			return err
```

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	return nil
+}
+
+func (r *dataFoundationClientReconciler) reconcileDefaultFSStorageClassClaim() error {
+	r.Log.Info("Reconciling Default Filesystem StorageClassClaim")
+
+	_, err := r.CreateOrUpdate(&r.defaultFSStorageClassClaim, func() error {
+		if err := r.own(&r.defaultFSStorageClassClaim, true); err != nil {
+			return fmt.Errorf("unable to set ownerRef on Default Filesystem StorageClassClaim: %v", err)
```
nb-ohad, line 227: Same as above
```suggestion
			return err
```

nb-ohad (changes requested):

cmd/crdCreator/main.go
```diff
@@ -41,6 +41,7 @@ func main() {
 		"noobaas.noobaa.io.yaml",
 		"objectbucketclaims.objectbucket.io.yaml",
 		"objectbuckets.objectbucket.io.yaml",
+		"storageclients.ocs.openshift.io.yaml",
```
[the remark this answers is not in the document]
leelavg: iirc, these are required for registering with scheme even before offering CR is created.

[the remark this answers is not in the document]
bindrad: We need them because we are trying to watch these resources and definitions for these resources are not available in the cluster at the time of registering watches.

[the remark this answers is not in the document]
nb-ohad: But we had a fix for that. What happened? 
The issue is that by introducing CRD ourselves, OLM will not override them when installing the offering package. This was only supposed to be used for CRDs that the offerings watch but NOT install.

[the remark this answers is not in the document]
bindrad: > But we had a fix for that. What happened?
Are you talking about adding dynamic watches?

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,241 @@
...
+	r.ManagedFusionOfferingReconciler = reconciler
+	r.offering = offering
+
+	r.storageClient.Name = "ocs-storageclient"
+	r.storageClient.Namespace = offering.Namespace
+
+	r.defaultBlockStorageClassClaim.Name = "default-ocs-block-storageclass-claim"
+	r.defaultBlockStorageClassClaim.Namespace = offering.Namespace
+
+	r.defaultFSStorageClassClaim.Name = "default-ocs-fs-storageclass-claim"
```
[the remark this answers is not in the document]
ezio-auditore: Changed to the appropriate one `ocs-storagecluster-cephfs`

controllers/datafoundation_client.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+func (r *dataFoundationClientReconciler) initReconciler(reconciler *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) {
+	r.ManagedFusionOfferingReconciler = reconciler
+	r.offering = offering
+
+	r.storageClient.Name = "ocs-storageclient"
+	r.storageClient.Namespace = offering.Namespace
+
+	r.defaultBlockStorageClassClaim.Name = "default-ocs-block-storageclass-claim"
```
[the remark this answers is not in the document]
ezio-auditore: changed to appropriate one `ocs-storagecluster-ceph-rbd`

controllers/datafoundation_client.go
ezio-auditore: Moved to reconcile func

controllers/datafoundation_client.go
ezio-auditore: updated

controllers/datafoundation_client.go
ezio-auditore: updated

controllers/datafoundation_client.go
ezio-auditore: updated

controllers/datafoundation_client.go
ezio-auditore: updated

controllers/datafoundation_client.go
ezio-auditore: updated

controllers/datafoundation_client.go
ezio-auditore: updated

controllers/datafoundation_client.go
ezio-auditore: upadated

controllers/datafoundation_client.go
ezio-auditore: updated

controllers/datafoundation_client.go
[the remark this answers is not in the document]
ezio-auditore: Updated

controllers/datafoundation_client.go
ezio-auditore: updated

cmd/crdCreator/main.go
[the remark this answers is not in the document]
nb-ohad: no, about the op restart solution

controllers/datafoundation_client.go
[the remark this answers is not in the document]
nb-ohad: Are we sure these are the names for provider mode in 4.11. 
If I remember correctly this is the name for the converged mode default block storage class.
I would suggest to talk with @Madhu-1 about this

controllers/datafoundation_client.go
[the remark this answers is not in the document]
nb-ohad: Are we sure these are the names for provider mode in 4.11. 
If I remember correctly this is the name for the converged mode default fs storage class.
I would suggest talking with @Madhu-1 about this

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,226 @@
...
+func (r *dataFoundationClientReconciler) reconcileStorageClient() error {
+	r.Log.Info("Reconciling StorageClient")
+
+	if r.dataFoundationClientSpec.providerEndpoint == "" {
+		return fmt.Errorf("dataFoundationProviderEndpoint is empty")
+	}
+
+	dfStorageProviderEndpointAsURL, err := url.ParseRequestURI(r.dataFoundationClientSpec.providerEndpoint)
+	if err != nil {
+		return err
```
nb-ohad, line 147: Please add context to this error as the error will be a generic URL parse failure error

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+func (r *dataFoundationClientReconciler) reconcileStorageClient() error {
+	r.Log.Info("Reconciling StorageClient")
+
+	if r.dataFoundationClientSpec.providerEndpoint == "" {
+		return fmt.Errorf("dataFoundationProviderEndpoint is empty")
+	}
+
+	dfStorageProviderEndpointAsURL, err := url.ParseRequestURI(r.dataFoundationClientSpec.providerEndpoint)
```
nb-ohad, line 145: No need to specify the type (`asURL`, and by the why this does not parse a URL it parses a URI) if this is the target type.
```suggestion
	providerEndpoint, err := url.ParseRequestURI(r.dataFoundationClientSpec.providerEndpoint)
```

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,226 @@
...
+	if err != nil {
+		return err
+	}
+	if dfStorageProviderEndpointAsURL.Host == "" {
```
nb-ohad, line 149:
```suggestion
	if providerEndpoint.Host == "" {
```

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,226 @@
...
+		}
+		return ctrl.Result{}, nil
+	}
+}
+
+func (r *dataFoundationClientReconciler) reconcileStorageClient() error {
+	r.Log.Info("Reconciling StorageClient")
+
+	if r.dataFoundationClientSpec.providerEndpoint == "" {
+		return fmt.Errorf("dataFoundationProviderEndpoint is empty")
```
nb-ohad, line 142: A field can be empty but not a value. There are two options here:
```suggestion
		return fmt.Errorf("dataFoundationClientSpec.providerEndpoint is empty")
```

Or
```suggestion
		return fmt.Errorf("Invalid provider endpoint, empty string")
```

I prefer the second option

cmd/crdCreator/main.go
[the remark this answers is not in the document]
bindrad: We didn't do that and decided to install CRDs on our own using the init container.

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	dfStorageProviderEndpointAsURL, err := url.ParseRequestURI(r.dataFoundationClientSpec.providerEndpoint)
+	if err != nil {
+		return err
+	}
+	if dfStorageProviderEndpointAsURL.Host == "" {
+		return fmt.Errorf("cannot retrieve host from %s, might be a malformed or partial URL", r.dataFoundationClientSpec.providerEndpoint)
```
nb-ohad, line 150: An error text should not start with a verb 
```suggestion
		return fmt.Errorf("Invalid provider endpoint %s, URL does not contain host", r.dataFoundationClientSpec.providerEndpoint)
```

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	if r.dataFoundationClientSpec.onboardingTicket == "" {
+		return fmt.Errorf("onboardingTicket is empty")
```
nb-ohad, line 154:
```suggestion
		return fmt.Errorf("Invalid onboarding ticket, empty string")
```

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,226 @@
...
+
+func (r *dataFoundationClientReconciler) reconcileCSV() error {
+	r.Log.Info("Reconciling OCS Client Operator CSV")
+
+	csv, err := r.getCSVByPrefix(ocsClientOperatorName)
+	if err != nil {
+		return fmt.Errorf("unable to get OCS client CSV: %v", err)
+	}
+	if err := r.own(csv, false); err != nil {
+		return err
```
nb-ohad, line 179: These statements are no longer wrapped in an error `CreateOrUpdate` or `GetAndUpdate`. This means that the error itself will not be wrapped with another error the brings the context. Please add context
```suggestion
		return fmt.Errorf("unable to set owner reference on OCS client csv : %v", err)
```

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,226 @@
...
+
+func (r *dataFoundationClientReconciler) reconcileDefaultBlockStorageClassClaim() error {
+	r.Log.Info("Reconciling Default BlockPool StorageClassClaim")
+
+	_, err := r.CreateOrUpdate(&r.defaultBlockStorageClassClaim, func() error {
+		if err := r.own(&r.defaultBlockStorageClassClaim, true); err != nil {
+			return err
+		}
+		r.defaultBlockStorageClassClaim.Spec.Type = "blockpool"
+		r.defaultBlockStorageClassClaim.Spec.StorageClient = &ocsclient.StorageClientNamespacedName{
```
nb-ohad, line 195: Why are we using a custom struct here? why not just the regular `client.NamespacedName{}`?

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,226 @@
...
+
+func (r *dataFoundationClientReconciler) reconcileDefaultFSStorageClassClaim() error {
+	r.Log.Info("Reconciling Default Filesystem StorageClassClaim")
+
+	_, err := r.CreateOrUpdate(&r.defaultFSStorageClassClaim, func() error {
+		if err := r.own(&r.defaultFSStorageClassClaim, true); err != nil {
+			return err
+		}
+		r.defaultFSStorageClassClaim.Spec.Type = "sharedfilesystem"
+		r.defaultFSStorageClassClaim.Spec.StorageClient = &ocsclient.StorageClientNamespacedName{
```
nb-ohad, line 215: Same as the last comment

nb-ohad (changes requested):

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,226 @@
...
+
+func (r *dataFoundationClientReconciler) initReconciler(reconciler *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) {
+	r.ManagedFusionOfferingReconciler = reconciler
+	r.offering = offering
+
+	r.storageClient.Name = "ocs-storageclient"
+	r.storageClient.Namespace = offering.Namespace
+
+	r.defaultBlockStorageClassClaim.Name = "ocs-storagecluster-ceph-rbd"
+	r.defaultBlockStorageClassClaim.Namespace = offering.Namespace
```
Madhu-1, line 78: claim is clusterscoped resource

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,226 @@
...
+
+	r.defaultFSStorageClassClaim.Name = "ocs-storagecluster-cephfs"
+	r.defaultFSStorageClassClaim.Namespace = offering.Namespace
```
Madhu-1, line 81: claim is clusterscoped resource

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,226 @@
...
+	return nil
+}
+
+func (r *dataFoundationClientReconciler) reconcileDefaultBlockStorageClassClaim() error {
+	r.Log.Info("Reconciling Default BlockPool StorageClassClaim")
+
+	_, err := r.CreateOrUpdate(&r.defaultBlockStorageClassClaim, func() error {
+		if err := r.own(&r.defaultBlockStorageClassClaim, true); err != nil {
+			return err
+		}
```
Madhu-1, line 193: you are trying to own clusterscoped resource from namespace resource i dont think it will work

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,226 @@
...
+		return fmt.Errorf("unable to create/update Default blockpool StorageClassClaim: %v", err)
+	}
+	return nil
+}
+
+func (r *dataFoundationClientReconciler) reconcileDefaultFSStorageClassClaim() error {
+	r.Log.Info("Reconciling Default Filesystem StorageClassClaim")
+
+	_, err := r.CreateOrUpdate(&r.defaultFSStorageClassClaim, func() error {
+		if err := r.own(&r.defaultFSStorageClassClaim, true); err != nil {
```
Madhu-1, line 211: you are trying to own clusterscoped resource from namespace resource i dont think it will work

controllers/datafoundationclient.go
nb-ohad: Yep, that will not work. 
We will need to be more careful in the uninstallation flow to make sure that we remove these as there is no direct ownership here.

bindrad:
@ezio-auditore IIRC, we can consume storage from multiple provider clusters on the application clusters. To do this we will need to create multiple ManagedFusionOffering CR with the correct config but this will result in issues.

- Multiple subscription will get created for client operator whereas only a single instance of client operator is required as it is a cluster-scoped operator
- The name of default storageClassClaim will be the same for all the connected providers which will create a conflict

controllers/datafoundationclient.go
ezio-auditore: That is what the ocs-client-operator api provides
https://github.com/red-hat-storage/ocs-client-operator/blob/ac2c276878b87322796c265c19644b44a4fe1459/api/v1alpha1/storageclassclaim_types.go#L48

controllers/datafoundation_client.go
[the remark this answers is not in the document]
rewantsoni: @nb-ohad I have a backup of storageClassClaim from consumer side and the name for blockpool is ocs-storagecluster-ceph-rbd and for sharedfilesystem is ocs-storagecluster-cephfs. For provider, we have the storageClassClaim name auto generated based on storageClassClaim name on provider side and the storageConsumerUID

controllers/datafoundation_client.go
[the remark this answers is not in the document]
rewantsoni: @nb-ohad I have a backup of storageClassClaim from consumer side and the name for blockpool is ocs-storagecluster-ceph-rbd and for sharedfilesystem is ocs-storagecluster-cephfs. For provider, we have the storageClassClaim name auto generated based on storageClassClaim name on provider side and the storageConsumerUID

nb-ohad:
@bindrad 
> Multiple subscription will get created for client operator whereas only a single instance of client operator is required as it is a cluster-scoped operator

That is wrong! you will always have one client operator but multiple storageclient CRs 

> The name of default storageClassClaim will be the same for all the connected providers which will create a conflict
By definition, there cannot be multiple default storage class claims, only for the first client and even these can be deleted.

@ezio-auditore please make sure you understand my answer on this.

controllers/datafoundationclient.go
Madhu-1: client.NamespacedName doesn't have the json tag which is required for unmarshal

controllers/datafoundationclient.go
ezio-auditore: Updated

controllers/datafoundationclient.go
ezio-auditore: Updated

controllers/datafoundationclient.go
ezio-auditore: Updated

controllers/datafoundationclient.go
ezio-auditore: Updated

controllers/datafoundationclient.go
ezio-auditore: Added the error context

bindrad:
> That is wrong! you will always have one client operator but multiple storageclient CRs

@nb-ohad I know that we need one client operator and multiple storage client CRs but that's not the case based on the implementation of the DFC offering.
According to the current implementation, we will need to create multiple ManagedFusionOffering CR if we want to connect with multiple providers. If we are creating multiple ManagaedFusionOffering CR then Agent will try to create multiple subscriptions.

controllers/datafoundationclient.go
ezio-auditore: Removed owning the storageclassclaim

controllers/datafoundationclient.go
ezio-auditore: Removed owning the storageclassclaim

ezio-auditore:
> > The name of default storageClassClaim will be the same for all the connected providers which will create a conflict
> > By definition, there cannot be multiple default storage class claims, only for the first client and even these can be deleted.
> 
> @ezio-auditore please make sure you understand my answer on this.
@nb-ohad are targeting multiple providers for one consumer in this PR?

cmd/crdCreator/main.go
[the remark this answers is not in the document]
nb-ohad: @bindrad I will approve this for this PR. But please find someone to work on this ASAP. 
We cannot ship with CRs that we are placing instead of the offering themselves. And the alternative will need time for testing

controllers/datafoundation_client.go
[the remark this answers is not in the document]
nb-ohad: Thank you @rewantsoni for the clarification. The important thing is that a migrated consumer and a new deployment will look and behave the same.

controllers/datafoundation_client.go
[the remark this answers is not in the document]
nb-ohad: Thank you @rewantsoni for the clarification. The important thing is that a migrated consumer and a new deployment will look and behave the same.

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,217 @@
...
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+const (
+	ocsClientOperatorName = "ocs-client-operator"
+)
```
nb-ohad, line 21: Please move consts and vars section below type definitions.

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,217 @@
...
+
+func (r *dataFoundationClientReconciler) initReconciler(reconciler *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) {
+	r.ManagedFusionOfferingReconciler = reconciler
+	r.offering = offering
+
+	r.storageClient.Name = "ocs-storageclient"
+	r.storageClient.Namespace = offering.Namespace
+
+	r.defaultBlockStorageClassClaim.Name = "ocs-storagecluster-ceph-rbd"
+	r.defaultFSStorageClassClaim.Name = "ocs-storagecluster-cephfs"
```
nb-ohad, line 78: To keep this function consistent, let's leave an empty line between int code for different resources
```suggestion
	r.defaultBlockStorageClassClaim.Name = "ocs-storagecluster-ceph-rbd"
	
	r.defaultFSStorageClassClaim.Name = "ocs-storagecluster-cephfs"
```

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,217 @@
...
+
+	providerEndpoint, found := offering.Spec.Config["dataFoundationProviderEndpoint"]
+	if !found {
+		r.Log.Error(
+			fmt.Errorf("missing field: dataFoundationProviderEndpoint"),
+			"an error occurred while parsing ManagedFusionOffering Data Foundation Client spec",
+		)
+		isValid = false
+	}
+	if !isValid {
```
nb-ohad, line 102: You have a space between checks of different fields so it makes sense to have a space here as well
```suggestion
	}
	
	if !isValid {
```

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,217 @@
...
+		}
+		return ctrl.Result{}, nil
+	}
+}
+
+func (r *dataFoundationClientReconciler) reconcileStorageClient() error {
+	r.Log.Info("Reconciling StorageClient")
+
+	if r.dataFoundationClientSpec.providerEndpoint == "" {
+		return fmt.Errorf("Invalid provider endpoint, empty string")
```
nb-ohad, line 139: Errors should start with a lower case letter. @ezio-auditore please configure your work environment to highlight such discrepancies.

```suggestion
		return fmt.Errorf("invalid provider endpoint, empty string")
```

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	providerEndpoint, err := url.ParseRequestURI(r.dataFoundationClientSpec.providerEndpoint)
+	if err != nil {
+		return fmt.Errorf("Failed to parse %s URI", r.dataFoundationClientSpec.providerEndpoint)
```
nb-ohad, line 144: Lower case at the start of an error, and also attach the original error.
```suggestion
		return fmt.Errorf("failed to parse %s, %v", r.dataFoundationClientSpec.providerEndpoint, err)
```

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+	if providerEndpoint.Host == "" {
+		return fmt.Errorf("Invalid provider endpoint %s, URL does not contain host", r.dataFoundationClientSpec.providerEndpoint)
```
nb-ohad, line 147: - Lower case at the start of an error
-  cannot be a URL one time and a URI another time. Let's remove the designation completely. 
```suggestion
		return fmt.Errorf("invalid provider endpoint %s, does not contain host", r.dataFoundationClientSpec.providerEndpoint)
```

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	if r.dataFoundationClientSpec.onboardingTicket == "" {
+		return fmt.Errorf("Invalid onboarding ticket, empty string")
```
nb-ohad, line 151: Lower case at the start of an error
```suggestion
		return fmt.Errorf("invalid onboarding ticket, empty string")
```

nb-ohad (changes requested):

nb-ohad:
@ezio-auditore Are you planning in adding offering removal code as a separate PR?

controllers/datafoundationclient.go

controllers/datafoundationclient.go

controllers/datafoundationclient.go

controllers/datafoundationclient.go

controllers/datafoundationclient.go

controllers/datafoundationclient.go

controllers/datafoundationclient.go

ezio-auditore:
> @ezio-auditore Are you planning to add the offering's removal code as a separate PR?

Yes, Uninstallation will be in a seperate PR.

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,219 @@
...
+		onboardingTicket: onboardingTicket,
+		providerEndpoint: providerEndpoint,
+	}
+	return nil
+}
+
+func (r *dataFoundationClientReconciler) reconcilePhases() (ctrl.Result, error) {
+	if !r.offering.DeletionTimestamp.IsZero() {
+		return ctrl.Result{}, nil
+	} else {
```
Madhu-1, line 120: as you are returning inside `if` you can skip `else` here

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,219 @@
...
+		}
+		if err := r.reconcileCSV(); err != nil {
+			return ctrl.Result{}, err
+		}
+		if err := r.reconcileDefaultBlockStorageClassClaim(); err != nil {
+			return ctrl.Result{}, err
+		}
+		if err := r.reconcileDefaultFSStorageClassClaim(); err != nil {
+			return ctrl.Result{}, err
+		}
```
Madhu-1, line 132: the oder should be as below isnt it?

```
reconcileCSV
reconcileStorageClient
reconcileDefaultStorageClassClaim
```

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,219 @@
...
+
+	r.storageClient.Name = "ocs-storageclient"
+	r.storageClient.Namespace = offering.Namespace
+
+	r.defaultBlockStorageClassClaim.Name = "ocs-storagecluster-ceph-rbd"
+
+	r.defaultFSStorageClassClaim.Name = "ocs-storagecluster-cephfs"
+}
+
+func (r *dataFoundationClientReconciler) parseSpec(offering *v1alpha1.ManagedFusionOffering) error {
```
Madhu-1, line 82: Good to add unit testing for the independent block of code to ensure it works as expected

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,219 @@
...
+			return ctrl.Result{}, err
+		}
+		if err := r.reconcileDefaultFSStorageClassClaim(); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+}
+
+func (r *dataFoundationClientReconciler) reconcileStorageClient() error {
```
Madhu-1, line 137: Good to add unit testing for the independent block of code to ensure it works as expected

controllers/datafoundationclient.go
```diff
@@ -0,0 +1,219 @@
...
+
+func (r *dataFoundationClientReconciler) reconcileCSV() error {
+	r.Log.Info("Reconciling OCS Client Operator CSV")
+
+	csv, err := r.getCSVByPrefix(ocsClientOperatorName)
+	if err != nil {
+		return fmt.Errorf("unable to get OCS client CSV: %v", err)
+	}
+	if err := r.own(csv, false); err != nil {
+		return fmt.Errorf("unable to set owner reference on OCS client csv : %v", err)
```
Madhu-1, line 178:
```suggestion
                return fmt.Errorf("unable to set owner reference on OCS client CSV: %v", err)
```

controllers/managedfusionoffering_controller.go
```diff
@@ -320,12 +320,23 @@ func pluginIsReadyToBeRemoved(reconciler *ManagedFusionOfferingReconciler, offer
...
+	switch offering.Spec.Kind {
+	case v1alpha1.KindDataFoundation:
+		if _, err := dfReconcile(reconciler, offering); err != nil {
+			return ctrl.Result{}, err
+		}
+	case v1alpha1.KindDataFoundationClient:
+		if _, err := dfcReconcile(reconciler, offering); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
```
Madhu-1, line 332: the default case is missing here?

controllers/managedfusionoffering_controller.go
```diff
@@ -320,12 +320,23 @@ func pluginIsReadyToBeRemoved(reconciler *ManagedFusionOfferingReconciler, offer
...
 func pluginReconcile(reconciler *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) (ctrl.Result, error) {
-	return dfReconcile(reconciler, offering)
+	switch offering.Spec.Kind {
+	case v1alpha1.KindDataFoundation:
+		if _, err := dfReconcile(reconciler, offering); err != nil {
+			return ctrl.Result{}, err
+		}
+	case v1alpha1.KindDataFoundationClient:
+		if _, err := dfcReconcile(reconciler, offering); err != nil {
+			return ctrl.Result{}, err
```
Madhu-1, line 330: `dfcReconcile` is already returning ctrl.Result, why not use it? if its not required dont return it

Madhu-1 (commented):
Left some suggestions/questions. LGTM

controllers/datafoundationclient.go
nb-ohad: On a general note I agree, but our in-repo test strategy is different. We are testing the entire reconcile operation as a whole, given different conditions expecting specific results.

controllers/managedfusionoffering_controller.go
nb-ohad: @Madhu-1 is right, the way that it is written wright now is actually a bug

ezio-auditore: updated

nb-ohad (approved):

PR #31, opened 2023-03-27
Add reconcileEgress policy for OVN and SDN CNI

bindrad:
Egress policies were missing so this PR introduces egress policies for OVN and SDN CNI based on CRD availability in the cluster.

### DIFF
--- a/controllers/managedfusion_controller.go
+++ b/controllers/managedfusion_controller.go
@@ -49,6 +49,9 @@ import (
 )
 
 const (
+	EgressFirewallCRD      = "egressfirewalls.k8s.ovn.org"
+	EgressNetworkPolicyCRD = "egressnetworkpolicies.network.openshift.io"
+
 	managedFusionFinalizer           = "managedfusion.ibm.com/finalizer"
 	managedFusionSecretName          = "managed-fusion-agent-config"
 	prometheusName                   = "managed-fusion-prometheus"
@@ -73,6 +76,9 @@ type ManagedFusionReconciler struct {
 	Log                          logr.Logger
 	Scheme                       *runtime.Scheme
 	Namespace                    string
+	AvailableCRDs                map[string]bool
+	CustomerNotificationHTMLPath string
+
 	ctx                          context.Context
 	prometheus                   *promv1.Prometheus
 	prometheusKubeRBACConfigMap  *corev1.ConfigMap
@@ -86,7 +92,6 @@ type ManagedFusionReconciler struct {
 	managedFusionSecret          *corev1.Secret
 	smtpConfigData               *smtpConfig
 	pagerDutyConfigData          *pagerDutyConfig
-	CustomerNotificationHTMLPath string
 }
 
 type smtpConfig struct {
--- a/controllers/managedfusion_controller.go
+++ b/controllers/managedfusion_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"strings"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -41,6 +42,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	"github.com/go-logr/logr"
+	openshiftv1 "github.com/openshift/api/network/v1"
+	ovnv1 "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/egressfirewall/v1"
 	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	promv1a1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1alpha1"
 	"github.com/red-hat-storage/managed-fusion-agent/templates"
@@ -68,6 +71,8 @@ const (
 	monLabelValue                    = "managed-fusion-agent"
 	amSecretPagerDutyServiceKeyKey   = "pagerDutyServiceKey"
 	amSecretSMTPAuthPasswordKey      = "smtpAuthPassword"
+	egressNetworkPolicyName          = "egress-rule"
+	egressFirewallName               = "default"
 )
 
 type ManagedFusionReconciler struct {
@@ -92,6 +97,8 @@ type ManagedFusionReconciler struct {
 	managedFusionSecret          *corev1.Secret
 	smtpConfigData               *smtpConfig
 	pagerDutyConfigData          *pagerDutyConfig
+	egressNetworkPolicy          *openshiftv1.EgressNetworkPolicy
+	egressFirewall               *ovnv1.EgressFirewall
 }
 
 type smtpConfig struct {
@@ -159,7 +166,7 @@ func (r *ManagedFusionReconciler) SetupWithManager(mgr ctrl.Manager, ctrlOptions
 			}}
 		},
 	)
-	return ctrl.NewControllerManagedBy(mgr).
+	ctrlBuilder := ctrl.NewControllerManagedBy(mgr).
 		WithOptions(*ctrlOptions).
 		For(&corev1.Secret{}, managedFusionSecretPredicates).
 		// Watch owned resources
@@ -185,8 +192,15 @@ func (r *ManagedFusionReconciler) SetupWithManager(mgr ctrl.Manager, ctrlOptions
 			&source.Kind{Type: &promv1.PrometheusRule{}},
 			enqueueManagedFusionAgentRequest,
 			monResourcesPredicates,
-		).
-		Complete(r)
+		)
+
+	if r.AvailableCRDs[EgressFirewallCRD] {
+		ctrlBuilder = ctrlBuilder.Owns(&ovnv1.EgressFirewall{})
+	}
+	if r.AvailableCRDs[EgressNetworkPolicyCRD] {
+		ctrlBuilder = ctrlBuilder.Owns(&openshiftv1.EgressNetworkPolicy{})
+	}
+	return ctrlBuilder.Complete(r)
 }
 
 // Reconcile changes to all owned resource based on the infromation provided by the ManangedFusion secret resource
@@ -269,6 +283,14 @@ func (r *ManagedFusionReconciler) initReconciler(ctx context.Context, req ctrl.R
 	r.alertRelabelConfigSecret.Name = alertRelabelConfigSecretName
 	r.alertRelabelConfigSecret.Namespace = r.Namespace
 
+	r.egressNetworkPolicy = &openshiftv1.EgressNetworkPolicy{}
+	r.egressNetworkPolicy.Name = egressNetworkPolicyName
+	r.egressNetworkPolicy.Namespace = r.Namespace
+
+	r.egressFirewall = &ovnv1.EgressFirewall{}
+	r.egressFirewall.Name = egressFirewallName
+	r.egressFirewall.Namespace = r.Namespace
+
 	r.smtpConfigData = &smtpConfig{}
 	r.pagerDutyConfigData = &pagerDutyConfig{}
 }
@@ -327,6 +349,13 @@ func (r *ManagedFusionReconciler) reconcilePhases() (reconcile.Result, error) {
 		if err := r.reconcilePrometheusProxyNetworkPolicy(); err != nil {
 			return ctrl.Result{}, err
 		}
+		if err := r.reconcileEgressFirewall(); err != nil {
+			return ctrl.Result{}, err
+		}
+		if err := r.reconcileEgressNetworkPolicy(); err != nil {
+			return ctrl.Result{}, err
+		}
+
 	}
 
 	return ctrl.Result{}, nil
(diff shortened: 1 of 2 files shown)


controllers/managedfusion_controller.go
```diff
@@ -689,6 +723,158 @@ func (r *ManagedFusionReconciler) reconcilePrometheusProxyNetworkPolicy() error
...
+		smtpEgressRule.Type = openshiftv1.EgressNetworkPolicyRuleAllow
+
+		desired.Spec.Egress = append(
+			[]openshiftv1.EgressNetworkPolicyRule{
+				smtpEgressRule,
+			},
+			desired.Spec.Egress...,
+		)
+		r.egressNetworkPolicy.Spec = desired.Spec
+		return nil
```
nb-ohad, line 870: This is the exact same code as the prev func but with slightly different struct names. 
Having this duplication could lead to some issues solved only on one func and not the other. 

Lets have a `getEgressDesiredState` function that returns `*EgressFirewallSpec`. In the first func we just call it and set it and in this one we call it and translate the desired state from `EgressFirewallSpec` to  `EgressNetworkPolicySpec`

controllers/managedfusion_controller.go
```diff
@@ -689,6 +723,158 @@ func (r *ManagedFusionReconciler) reconcilePrometheusProxyNetworkPolicy() error
...
+		// Inserting the VPC Egress rule in front of all other egress rules.
+		// The order or rules matter
+		// https://docs.openshift.com/container-platform/4.10/networking/openshift_sdn/configuring-egress-firewall.html#policy-rule-order_openshift-sdn-egress-firewall
+		// Inserting this rule in the front ensures it comes before the EgressNetworkPolicyRuleDeny rule.
+		desired.Spec.Egress = append(
+			vpcEgressRules,
+			desired.Spec.Egress...,
+		)
+
+		smtpURL, err := url.Parse(r.smtpConfigData.Endpoint)
```
leelavg, line 775: - if the endpoint should strictly be a url with protocol included ex: `http://example.com` vs `example.com`, it's better to use `url.ParseRequestURI` rather than `url.Parse`
- In general `url.Parse` is very loosely validated

controllers/managedfusion_controller.go
```diff
@@ -689,6 +723,158 @@ func (r *ManagedFusionReconciler) reconcilePrometheusProxyNetworkPolicy() error
...
```
leelavg, line 851: same as above.

leelavg (commented):
- Is SDN still a hard requirement?

nb-ohad (changes requested):

bindrad:
> * Is SDN still a hard requirement?

Yes, it is a requirement because the existing consumer clusters that will be migrated to the new service will be still using SDN.

controllers/managedfusion_controller.go
bindrad: It will be a lot of type conversion, let me present something

controllers/managedfusion_controller.go
bindrad: Updated

controllers/managedfusion_controller.go
bindrad: Updated

controllers/managedfusion_controller.go
```diff
@@ -689,6 +723,158 @@ func (r *ManagedFusionReconciler) reconcilePrometheusProxyNetworkPolicy() error
...
+	}
+
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.egressFirewall, func() error {
+		if err := r.own(r.egressFirewall); err != nil {
+			return err
+		}
+		desired := templates.EgressFirewallTemplate.DeepCopy()
+
+		// Get the aws config map
+		awsConfigMap := &corev1.ConfigMap{}
```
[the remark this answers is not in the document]
bindrad: Updated

controllers/managedfusion_controller.go
```diff
@@ -689,6 +723,158 @@ func (r *ManagedFusionReconciler) reconcilePrometheusProxyNetworkPolicy() error
...
+		awsConfigMap.Name = utils.IMDSConfigMapName
+		awsConfigMap.Namespace = r.Namespace
+		if err := r.get(awsConfigMap); err != nil {
+			return fmt.Errorf("unable to get AWS ConfigMap: %v", err)
```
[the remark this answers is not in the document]
bindrad: Updated

controllers/managedfusion_controller.go
```diff
@@ -689,6 +723,158 @@ func (r *ManagedFusionReconciler) reconcilePrometheusProxyNetworkPolicy() error
...
+		}
+
+		// Get the machine cidr
+		vpcCIDR, ok := awsConfigMap.Data[utils.CIDRKey]
+		if !ok {
+			return fmt.Errorf("unable to determine machine CIDR from AWS ConfigMap")
```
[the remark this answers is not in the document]
bindrad: Updated

controllers/managedfusion_controller.go
bindrad: Updated

controllers/managedfusion_controller.go
```diff
@@ -689,6 +723,158 @@ func (r *ManagedFusionReconciler) reconcilePrometheusProxyNetworkPolicy() error
...
+		}
+
+		smtpEgressRule := ovnv1.EgressFirewallRule{}
+		smtpEgressRule.To.DNSName = smtpURL.Hostname()
+		smtpEgressRule.Type = ovnv1.EgressFirewallRuleAllow
+
+		desired.Spec.Egress = append(
+			[]ovnv1.EgressFirewallRule{
+				smtpEgressRule,
+			},
```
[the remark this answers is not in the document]
nb-ohad: The fact the we did something in a way that has very obvious cons just does not mean we need to repeat the mistake here. Let go forward with the requested change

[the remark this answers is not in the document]
nb-ohad: ok

controllers/managedfusion_controller.go
```diff
@@ -689,6 +723,123 @@ func (r *ManagedFusionReconciler) reconcilePrometheusProxyNetworkPolicy() error
...
+		var desiredEgressNetworkPolicy openshiftv1.EgressNetworkPolicy
+		for i := range desiredEgressFirewallSpec.Egress {
+			desiredEgressNetworkPolicy.Spec.Egress = append(
+				desiredEgressNetworkPolicy.Spec.Egress,
+				openshiftv1.EgressNetworkPolicyRule{
+					Type: openshiftv1.EgressNetworkPolicyRuleType(desiredEgressFirewallSpec.Egress[i].Type),
+					To:   openshiftv1.EgressNetworkPolicyPeer(desiredEgressFirewallSpec.Egress[i].To),
+				},
+			)
+		}
```
nb-ohad, line 772: The conversion here is problematic as it depends on implementation knowledge of the different structs and enum values

I would prefer to have an explicit conversion and I am suggesting the following util function (in the reconciler file, but not on the reconciler itself)

```
func egressFirewallSpecToEgressNetworkPolicySpec(firewallSpec *ovnv1.EgressFirewallSpec) *openshiftv1.EgressNetworkPolicySpec {
	networkPolicySpec := &openshiftv1.EgressNetworkPolicySpec{}

	for i := range firewallSpec.Egress {
		firewalRule := &firewallSpec.Egress[i]
		networkPolicyRule := openshiftv1.EgressNetworkPolicyRule{}

		// Convert rule type
		switch firewalRule.Type {
		case ovnv1.EgressFirewallRuleAllow:
			networkPolicyRule.Type = openshiftv1.EgressNetworkPolicyRuleAllow
		case ovnv1.EgressFirewallRuleDeny:
			networkPolicyRule.Type = openshiftv1.EgressNetworkPolicyRuleDeny
		}

		// Convert Rule Peer
		networkPolicyRule.To = openshiftv1.EgressNetworkPolicyPeer{
			CIDRSelector: firewalRule.To.CIDRSelector,
			DNSName:      firewalRule.To.DNSName,
		}

		networkPolicySpec.Egress = append(
			networkPolicySpec.Egress,
			networkPolicyRule,
		)
	}

	return networkPolicySpec
}

```

nb-ohad (changes requested):

controllers/managedfusion_controller.go
```diff
@@ -689,6 +723,123 @@ func (r *ManagedFusionReconciler) reconcilePrometheusProxyNetworkPolicy() error
...
+	imdsConfigMap.Name = utils.IMDSConfigMapName
+	imdsConfigMap.Namespace = r.Namespace
+	if err := r.get(imdsConfigMap); err != nil {
+		return nil, fmt.Errorf("unable to get AWS IMDS ConfigMap: %v", err)
+	}
+
+	// Get the machine cidr
+	vpcCIDR, ok := imdsConfigMap.Data[utils.CIDRKey]
+	if !ok {
+		return nil, fmt.Errorf("unable to determine machine CIDR from AWS ConfigMap")
```
nb-ohad, line 796: Still missing imds
```suggestion
		return nil, fmt.Errorf("unable to determine machine CIDR from AWS IMDS ConfigMap")
```

nb-ohad (changes requested):

controllers/managedfusion_controller.go
```diff
@@ -689,6 +723,123 @@ func (r *ManagedFusionReconciler) reconcilePrometheusProxyNetworkPolicy() error
...
+		r.egressNetworkPolicy.Spec = desiredEgressNetworkPolicy.Spec
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to update egressNetworkPolicy: %v", err)
+	}
+	return nil
+}
+
+func (r *ManagedFusionReconciler) getEgressDesiredState() (*ovnv1.EgressFirewallSpec, error) {
```
leelavg, line 782: - After looking through both CRDs why not return only the rules here?

``` go
type egressRules struct {
 type_ string
 dnsName string
 cidrSelector string
}

// callee signature be
func (r *ManagedFusionReconciler) getDesiredEgressRules() ([]egressRules, error)

// caller will just use either `dnsName` or `cidrSelector` whatever is set
```

(or)

- if having another struct is a no-go, let the callee fill the rules when the spec if supplied and switch based on available CRDs?

rationale being, the current implementation is trying to retrieve the info even though we were the one's who set the info in the first place.

controllers/managedfusion_controller.go
```diff
@@ -322,6 +349,13 @@ func (r *ManagedFusionReconciler) reconcilePhases() (reconcile.Result, error) {
 		if err := r.reconcilePrometheusProxyNetworkPolicy(); err != nil {
 			return ctrl.Result{}, err
 		}
+		if err := r.reconcileEgressFirewall(); err != nil {
+			return ctrl.Result{}, err
+		}
+		if err := r.reconcileEgressNetworkPolicy(); err != nil {
+			return ctrl.Result{}, err
+		}
```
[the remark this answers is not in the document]
bindrad: Yes, for now, these are limited to the agent namespace.
We will need a separate set of egress policies in the offering namespace and that will be decided based on offering type.

controllers/managedfusion_controller.go
```diff
@@ -154,7 +166,7 @@ func (r *ManagedFusionReconciler) SetupWithManager(mgr ctrl.Manager, ctrlOptions
 			}}
 		},
 	)
-	return ctrl.NewControllerManagedBy(mgr).
+	managedOCSController := ctrl.NewControllerManagedBy(mgr).
```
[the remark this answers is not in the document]
bindrad: Updated

controllers/managedfusion_controller.go
bindrad: Implemented the solution that @leelavg suggested

controllers/managedfusion_controller.go
```diff
@@ -689,6 +723,123 @@ func (r *ManagedFusionReconciler) reconcilePrometheusProxyNetworkPolicy() error
...
+	}
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.egressNetworkPolicy, func() error {
+		if err := r.own(r.egressNetworkPolicy); err != nil {
+			return err
+		}
+		desiredEgressFirewallSpec, err := r.getEgressDesiredState()
+		if err != nil {
+			return err
+		}
+		var desiredEgressNetworkPolicy openshiftv1.EgressNetworkPolicy
```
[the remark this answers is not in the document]
bindrad: Implemented the solution that @leelavg suggested

controllers/managedfusion_controller.go
bindrad: Updated

controllers/managedfusion_controller.go
bindrad: Can you check if the latest commit has the changes that you suggested?

controllers/managedfusion_controller.go
```diff
@@ -689,6 +723,135 @@ func (r *ManagedFusionReconciler) reconcilePrometheusProxyNetworkPolicy() error
...
+	if err != nil {
+		return fmt.Errorf("failed to update egressNetworkPolicy: %v", err)
+	}
+	return nil
+}
+
+type egressRules struct {
+	ruleType     string
+	dnsName      string
+	cidrSelector string
```
leelavg, line 805: - At usage site, pls make sure only one of these are set as these are mutually exclustive
- If an empty string is considered unset then we are fine

nb-ohad (changes requested):

controllers/managedfusion_controller.go
```diff
@@ -689,6 +723,153 @@ func (r *ManagedFusionReconciler) reconcilePrometheusProxyNetworkPolicy() error
...
+	}
+
+	// Inserting the VPC Egress rule in front of all other egress rules.
+	// The order or rules matter
+	// https://docs.openshift.com/container-platform/4.10/networking/openshift_sdn/configuring-egress-firewall.html#policy-rule-order_openshift-sdn-egress-firewall
+	// Inserting this rule in the front ensures it comes before the EgressNetworkPolicyRuleDeny rule.
+	allowEgressRules = append(
+		allowEgressRules,
+		vpcEgressRules...,
+	)
```
nb-ohad, line 821:
```suggestion
```

nb-ohad (changes requested):

controllers/managedfusion_controller.go
```diff
@@ -689,6 +723,135 @@ func (r *ManagedFusionReconciler) reconcilePrometheusProxyNetworkPolicy() error
...
+	}
+	if smtpURL.Hostname() == "" {
+		return nil, fmt.Errorf("smtp config does not contain a host entry")
+	}
+
+	smtpEgressRule := egressRules{}
+	smtpEgressRule.dnsName = smtpURL.Hostname()
+	smtpEgressRule.ruleType = string(ovnv1.EgressFirewallRuleAllow)
+
+	desiredRules = append([]egressRules{smtpEgressRule}, desiredRules...)
```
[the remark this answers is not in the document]
bindrad: Yes, I have updated the PR

controllers/managedfusion_controller.go
bindrad: This is not relevant anymore as we have another approach to solve the problem.

nb-ohad (approved):

PR #32, opened 2023-03-28
Implement agent removal logic and offering removal protection

nb-ohad:
- refactor finalizer addition and removal into utils
- add finalizer logic for ManageFusionOffering CR
- intreduce a new IsReadyToRemove offering callback
- add offering deletion logic when managed fusion config is marked for deletion
- lowercase error strings on some errors

### DIFF
--- a/cmd/awsDataGather/suite_test.go
+++ b/cmd/awsDataGather/suite_test.go
@@ -1,11 +1,9 @@
 /*
-
-
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
--- a/controllers/datafoundation.go
+++ b/controllers/datafoundation.go
@@ -70,6 +70,10 @@ func DFAddToScheme(scheme *runtime.Scheme) {
 	utilruntime.Must(ocsv1.AddToScheme(scheme))
 }
 
+func dfIsReadyToBeRemoved(r *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) (bool, error) {
+	return true, nil
+}
+
 func dfReconcile(offeringReconciler *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) error {
 	r := dataFoundationReconciler{}
 	r.initReconciler(offeringReconciler, offering)
@@ -298,7 +302,7 @@ func (r *dataFoundationReconciler) reconcileCephIngressNetworkPolicy() error {
 		return nil
 	})
 	if err != nil {
-		return fmt.Errorf("Failed to update ceph ingress NetworkPolicy: %v", err)
+		return fmt.Errorf("failed to update ceph ingress NetworkPolicy: %v", err)
 	}
 	return nil
 }
@@ -313,7 +317,7 @@ func (r *dataFoundationReconciler) reconcileProviderAPIServerNetworkPolicy() err
 		return nil
 	})
 	if err != nil {
-		return fmt.Errorf("Failed to update provider api server NetworkPolicy: %v", err)
+		return fmt.Errorf("failed to update provider api server NetworkPolicy: %v", err)
 	}
 
 	return nil
@@ -321,7 +325,7 @@ func (r *dataFoundationReconciler) reconcileProviderAPIServerNetworkPolicy() err
 
 func (r *dataFoundationReconciler) reconcileRookCephOperatorConfig() error {
 	if err := r.get(r.rookConfigMap); err != nil {
-		return fmt.Errorf("Failed to get Rook ConfigMap: %v", err)
+		return fmt.Errorf("failed to get Rook ConfigMap: %v", err)
 	}
 
 	cloneRookConfigMap := r.rookConfigMap.DeepCopy()
@@ -333,7 +337,7 @@ func (r *dataFoundationReconciler) reconcileRookCephOperatorConfig() error {
 	cloneRookConfigMap.Data["ROOK_CSI_ENABLE_RBD"] = "false"
 	if !equality.Semantic.DeepEqual(r.rookConfigMap, cloneRookConfigMap) {
 		if err := r.update(cloneRookConfigMap); err != nil {
-			return fmt.Errorf("Failed to update Rook ConfigMap: %v", err)
+			return fmt.Errorf("failed to update Rook ConfigMap: %v", err)
 		}
 	}
 
--- a/controllers/managedfusion_controller.go
+++ b/controllers/managedfusion_controller.go
@@ -46,6 +46,7 @@ import (
 	ovnv1 "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/egressfirewall/v1"
 	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	promv1a1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1alpha1"
+	v1alpha1 "github.com/red-hat-storage/managed-fusion-agent/api/v1alpha1"
 	"github.com/red-hat-storage/managed-fusion-agent/templates"
 	"github.com/red-hat-storage/managed-fusion-agent/utils"
 	netv1 "k8s.io/api/networking/v1"
@@ -55,7 +56,7 @@ const (
 	EgressFirewallCRD      = "egressfirewalls.k8s.ovn.org"
 	EgressNetworkPolicyCRD = "egressnetworkpolicies.network.openshift.io"
 
-	managedFusionFinalizer           = "managedfusion.ibm.com/finalizer"
+	managedFusionFinalizer           = "misf.ibm.com/finalizer"
 	managedFusionSecretName          = "managed-fusion-agent-config"
 	prometheusName                   = "managed-fusion-prometheus"
 	prometheusServiceName            = "prometheus"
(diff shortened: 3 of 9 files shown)


controllers/datafoundation.go
```diff
@@ -70,6 +70,10 @@ func DFAddToScheme(scheme *runtime.Scheme) {
 	utilruntime.Must(ocsv1.AddToScheme(scheme))
 }
 
+func dfIsReadyToBeRemoved(r *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) (bool, error) {
```
leelavg, line 73: - Better have verb at the start in func sig, `dfIs` vs `isDf`

utils/utils.go
```diff
@@ -51,8 +51,29 @@ func Remove(slice []string, s string) (result []string) {
...
+		time.Sleep(sleep)
+		log.Println("retrying after error:", err)
+	}
+	return fmt.Errorf("Failed after %d retries. Last error: %v", attempts, err)
+
+}
+
 // AddLabel add a label to a resource metadata
-func AddLabel(obj metav1.Object, key string, value string) {
+func AddLabel(obj client.Object, key string, value string) {
```
leelavg, line 76: - why `client.Object` if we are only going to touch `meta` on the object but not any `runtime`?

controllers/managedfusion_controller.go
```diff
@@ -466,10 +484,10 @@ func (r *ManagedFusionReconciler) reconcileAlertmanagerSecret() error {
 			return err
 		}
 		if r.pagerDutyConfigData.ServiceKey == "" {
-			return fmt.Errorf("Agent PagerDuty configuration does not contain serviceKey entry")
+			return fmt.Errorf("agent PagerDuty configuration does not contain serviceKey entry")
```
leelavg, line 487: - as we are touching these lines anyways, why not turn `fmt.Errorf` to `errors.New` if we aren't interpolating strings/errors?
- same for remaining

controllers/managedfusionoffering_controller.go
```diff
@@ -247,6 +275,10 @@ func (r *ManagedFusionOfferingReconciler) getCSVByPrefix(name string) (*opv1a1.C
 
 // All the below functions are placeholder for offering plugin integration
 
+func pluginIsReadyToBeRemoved(reconciler *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) (bool, error) {
```
leelavg, line 278: - better have verb at the start of func `pluginIs` vs `isPlugin`

controllers/managedfusion_controller.go
```diff
@@ -297,26 +298,39 @@ func (r *ManagedFusionReconciler) initReconciler(ctx context.Context, req ctrl.R
...
 
 	} else if r.managedFusionSecret.UID != "" {
-		if !utils.Contains(r.managedFusionSecret.GetFinalizers(), managedFusionFinalizer) {
+		if added := utils.AddFinalizer(r.managedFusionSecret, managedFusionFinalizer); added {
 			r.Log.V(-1).Info(fmt.Sprintf("finalizer missing on the %s secret resource, adding...", managedFusionSecretName))
-			r.managedFusionSecret.SetFinalizers(append(r.managedFusionSecret.GetFinalizers(), managedFusionFinalizer))
 			if err := r.update(r.managedFusionSecret); err != nil {
 				return ctrl.Result{}, fmt.Errorf("failed to update %s secret with finalizer: %v", managedFusionSecretName, err)
 			}
+			r.Log.Info("Adding finallizer to %s secret completed successfuly", managedFusionSecretName)
```
leelavg, line 333:
```suggestion
			r.Log.Info(fmt.Sprintf("Adding finallizer to %s secret completed successfuly", managedFusionSecretName))
```

controllers/datafoundation.go
nb-ohad: the `df` prefix  is our internal marker for all offering-related callbacks, this is going to change when we separate the DF code into its own package

utils/utils.go
nb-ohad: This is the new best practice for all `client` and `ctrlutil` functionality. They used to have segregation between functions that take `metav1.Object` and `runtime.Object` but then they introduced the `client.Object` which aggregates both. We are just following their footsteps

controllers/managedfusion_controller.go
nb-ohad: We can. But I find it easier to use `fmt.Errorf` for consistency and to ease changes were we will want to introduce variables into the message

controllers/managedfusionoffering_controller.go
nb-ohad: the `plugin` prefix is our internal marker for all plugin infrastructure layer callbacks, this is going to change when we introduce an actual plugin abstraction.

controllers/managedfusion_controller.go
nb-ohad: Good catch! Updated

controllers/datafoundation.go
leelavg: > this is going to change
- ack, the current functions always read as that something is ready to be removed 😅 
- fine for now, if this is slated for a refactor

controllers/managedfusion_controller.go
leelavg: - ack, I remember reading to use `errors.New` if there's no interpolation
- not that it gets in the same, so I don't feel any pressing need for the change

controllers/managedfusion_controller.go
```diff
@@ -297,26 +298,39 @@ func (r *ManagedFusionReconciler) initReconciler(ctx context.Context, req ctrl.R
...
+			r.Log.Info("ManagedFusionOffering CRs found, issuing delete")
+			offering := v1alpha1.ManagedFusionOffering{}
+			if err := r.deleteAllOf(&offering); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to issue delete for one or more ManagedFusionOffering CRs, %v", err)
+			} else {
+				r.Log.Info("ManagedFusionOffering CRs delete issued, requeuing the event")
+				return ctrl.Result{RequeueAfter: 10}, nil
+			}
+		} else {
+			r.Log.Info("No offering found, removing finalizer from %s secret", managedFusionSecretName)
```
bindrad, line 314: Use `fmt.Sprintf()` or pass key and value to `r.Log.Info()`
```suggestion
			r.Log.Info(fmt.Sprintf("No offering found, removing finalizer from %s secret", managedFusionSecretName))
```

controllers/managedfusion_controller.go
```diff
@@ -297,26 +298,39 @@ func (r *ManagedFusionReconciler) initReconciler(ctx context.Context, req ctrl.R
...
+			if removed := utils.RemoveFinalizer(r.managedFusionSecret, managedFusionFinalizer); removed {
+				if err := r.update(r.managedFusionSecret); err != nil {
+					return ctrl.Result{}, fmt.Errorf("failed to remove finalizer from %s secret: %v", managedFusionSecretName, err)
+				}
+				r.Log.Info("Removing finallizer from %s secret completed successfuly", managedFusionSecretName)
```
bindrad, line 319: Same as above, use `fmt.Sprintf`

controllers/managedfusion_controller.go
```diff
@@ -297,26 +298,39 @@ func (r *ManagedFusionReconciler) initReconciler(ctx context.Context, req ctrl.R
...
-			r.Log.Info("finallizer removed successfully")
 
 			if err := r.initiateAgentUninstallation(); err != nil {
 				return ctrl.Result{}, fmt.Errorf("failed to initiate agent uninstallation %v", err)
 			}
 		}
 
 	} else if r.managedFusionSecret.UID != "" {
-		if !utils.Contains(r.managedFusionSecret.GetFinalizers(), managedFusionFinalizer) {
+		if added := utils.AddFinalizer(r.managedFusionSecret, managedFusionFinalizer); added {
```
bindrad, line 328: Why can't we use the utils provided by controller-runtime instead of implementing our own?
https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.9.5/pkg/controller/controllerutil#AddFinalizer

controllers/managedfusion_controller.go
```diff
@@ -297,26 +298,39 @@ func (r *ManagedFusionReconciler) initReconciler(ctx context.Context, req ctrl.R
...
+			offering := v1alpha1.ManagedFusionOffering{}
+			if err := r.deleteAllOf(&offering); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to issue delete for one or more ManagedFusionOffering CRs, %v", err)
+			} else {
+				r.Log.Info("ManagedFusionOffering CRs delete issued, requeuing the event")
+				return ctrl.Result{RequeueAfter: 10}, nil
+			}
+		} else {
+			r.Log.Info("No offering found, removing finalizer from %s secret", managedFusionSecretName)
+			if removed := utils.RemoveFinalizer(r.managedFusionSecret, managedFusionFinalizer); removed {
```
bindrad, line 315: Why can't we use the utils provided by controller-runtime instead of implementing our own?
https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.9.5/pkg/controller/controllerutil#RemoveFinalizer

controllers/managedfusion_controller.go
```diff
@@ -669,8 +687,8 @@ func (r *ManagedFusionReconciler) reconcileMonitoringResources() error {
 	r.Log.Info("reconciling monitoring resources")
 
 	podMonitorList := promv1.PodMonitorList{}
-	if err := r.list(&podMonitorList); err != nil {
-		return fmt.Errorf("Could not list pod monitors: %v", err)
+	if err := r.list(&podMonitorList, client.InNamespace(r.Namespace)); err != nil {
```
bindrad, line 690: Why should we list podMonitors, serviceMonitors and promRules in single namespace only(the namespace where monitoring stack is deployed)?
We should add labels to the monitoring resources that are created by the offering as well.
Example: cephPromRules are created by ocs-operator so our prometheus should be able to read those rules and raise alerts, for this to happen, we should add label to the cephPromRules.

controllers/managedfusion_controller.go
```diff
@@ -681,8 +699,8 @@ func (r *ManagedFusionReconciler) reconcileMonitoringResources() error {
 	}
 
 	serviceMonitorList := promv1.ServiceMonitorList{}
-	if err := r.list(&serviceMonitorList); err != nil {
-		return fmt.Errorf("Could not list service monitors: %v", err)
+	if err := r.list(&serviceMonitorList, client.InNamespace(r.Namespace)); err != nil {
```
bindrad, line 702: Same as above

controllers/managedfusion_controller.go
```diff
@@ -693,8 +711,8 @@ func (r *ManagedFusionReconciler) reconcileMonitoringResources() error {
 	}
 
 	promRuleList := promv1.PrometheusRuleList{}
-	if err := r.list(&promRuleList); err != nil {
-		return fmt.Errorf("Could not list prometheus rules: %v", err)
+	if err := r.list(&promRuleList, client.InNamespace(r.Namespace)); err != nil {
```
bindrad, line 714: same as above

bindrad (changes requested):

controllers/managedfusion_controller.go
nb-ohad: Because their finalizer utils do not return an indication if the new finalizer already exists or added.
We can change our utils to use a controllerutils implementation but that results in an unneeded iteration on the finalizer list which is unnecessary for such a simple operation

controllers/managedfusion_controller.go
nb-ohad: Because their finalizer utils do not return an indication if the finalizer was found and removed or not.
We can change our utils to use a controllerutils implementation but that results in an unneeded iteration on the finalizer list which is unnecessary for such a simple operation

controllers/managedfusion_controller.go
nb-ohad: Thinking about it again, it does not make sense to have this here at all. Each offering might have its own way of producing/creating the monitoring resources and might even reconcile back any changes to them.
For DF, OCS operator is adding he ability to ask it to add a desired label to the monitoring resources. This means the code above is not needed after we start deploying with ocs-operator 4.12. In the meantime, I think this code should move into `datafoundation.go` code, but that is for another PR.

controllers/managedfusion_controller.go
nb-ohad: done

controllers/managedfusion_controller.go
nb-ohad: done

controllers/managedfusion_controller.go
nb-ohad: I have reverted the changes to be moved to `datafoundation.go` at a future Pr

controllers/managedfusion_controller.go
nb-ohad: I have reverted the changes to be moved to `datafoundation.go` at a future Pr

controllers/managedfusion_controller.go
nb-ohad: I have reverted the changes to be moved to `datafoundation.go` at a future Pr

utils/utils.go
nb-ohad: I don't remember how many operator SDK versions / kubebuilder versions ago they did the change where they refactor all of the APIs from `runtime.Object` and `meta.Object` to `client.Object` for their own controller utils. You can check the their history their release nodes for more information

cmd/awsDataGather/suite_test.go
ezio-auditore, line 1: This change is already in a different PR. Please remove unrelated changes in this PR

controllers/datafoundation.go
ezio-auditore, line 1: Lint changes should go into a seperate PR

ezio-auditore (changes requested):
Please add managedfusionoffering CRD to the shim/crds directory for the tests to pass.
Also i see multiple files with lint changes. please keep lint changes in a seperate PR

bindrad:
> Please add managedfusionoffering CRD to the shim/crds directory for the tests to pass.

We don't need to do this because it is taken care of from the config/crd folder. We need to add the agent package to the scheme in the test(https://github.com/red-hat-storage/managed-fusion-agent/blob/main/controllers/suite_test.go#L123).

controllers/datafoundation.go
nb-ohad: On a general note, there is no need to do a separate PR, we can have a separate commit. Saying that lint changes are automatic on my machine and every file which I open and save gets them. Without completely altering my setup it will be very hard and time-consuming which I don't think justifies the cost. Unless you have a suggestion for an easy way around it that does not include editing each file with a simple text editor, I would prefer not to do that. 

Also please note that these are not general lint changes, these are changes that are only applied to files touched by this PR

nb-ohad: A better approach would be to have a CI task that makes sure we are not contributing code that does not run through the full lint process. This will ensure that no one can contribute code that does not meet our code standards

nb-ohad: @ezio-auditore So I think we are in agreement that we need a future tuning to the CI, which will not be happing within this PR. Can this thread be resolved now?

ezio-auditore: Yes, but I am unable to resolve this. Something with the perms i guess. Please resolve this thread

cmd/awsDataGather/suite_test.go
ezio-auditore: This too can be resolved

bindrad (approved):

PR #33, opened 2023-03-28
Data Foundation Offering Removal logic

nb-ohad:
- Implement `IsReadyToBeRemoved` hook to indicate that `stoargecluster` does not longer exists
- Observe offering deletion timestamp in data foundation reconciler code  to initiate removal logic
- Wait for all OCS `storageconsumer` CRs to be removed before deletion of the `storagecluster` CR


controllers/datafoundation.go
```diff
@@ -33,21 +52,21 @@ type dataFoundationSpec struct {
...
-	cephIngressNetworkPolicy       *netv1.NetworkPolicy
-	providerAPIServerNetworkPolicy *netv1.NetworkPolicy
-	rookConfigMap                  *corev1.ConfigMap
-	ocsInitialization              *ocsv1.OCSInitialization
+	onboardingValidationKeySecret  corev1.Secret
+	storageCluster                 ocsv1.StorageCluster
+	cephIngressNetworkPolicy       netv1.NetworkPolicy
+	providerAPIServerNetworkPolicy netv1.NetworkPolicy
+	rookConfigMap                  corev1.ConfigMap
+	ocsInitialization              ocsv1.OCSInitialization
```
leelavg, line 62: - Is the signature changed only to have default init of these structs?
- Even though I don't see much issue with this, now at every call site of sub-reconciler we need to remember to send a pointer which kinda differs from our main reconciler
- To be consistent, I'd like to have these as pointers

controllers/datafoundation.go
```diff
@@ -70,73 +89,99 @@ func DFAddToScheme(scheme *runtime.Scheme) {
...
 	r.providerAPIServerNetworkPolicy.Name = "provider-api-server-rule"
 	r.providerAPIServerNetworkPolicy.Namespace = offering.Namespace
 
-	r.ocsInitialization = &ocsv1.OCSInitialization{}
 	r.ocsInitialization.Name = "ocsinit"
 	r.ocsInitialization.Namespace = offering.Namespace
 }
 
+func (r *dataFoundationReconciler) reconcilePhases() (ctrl.Result, error) {
+	if r.offering.DeletionTimestamp.IsZero() {
```
leelavg, line 149:
```suggestion
	if !r.offering.DeletionTimestamp.IsZero() {
```

leelavg (changes requested):

leelavg:
- reviewed only latest commit

controllers/datafoundation.go
nb-ohad: Using pointers where not needed and then allocating separate objects on the heap move memory management toward fragmentation. I know, that for a project like this, it does not affect anything much, but I still follow best practices when I can. Dynamic allocation into pointers should never be the default if a language allows you (there are languages like javascript and python which default to dynamic allocation and make it hard to avoid them)

Saying that I agree that we need consistency, and I am intending to have a PR that fixes the other controller. I just felt that changes to that controller should not be part of this PR.

controllers/datafoundation.go
nb-ohad: fixed

controllers/datafoundation.go
leelavg: > I am intending to have a PR that fixes the other controller
- then it's fine, I know it'll not be much of a diff and so want this to be consistent across controllers that live in the current proj

leelavg (approved):
approval is only for latest commit.

bindrad (approved):

Issue #27, opened 2023-03-24
OCS Operator expects NooBaa CRD to be present on the cluster


bindrad:
We cannot directly install OCS-Operator because it expects NooBaa CRDs to be present on the cluster. Installing OCS-Operator does not bring NooBaa CRDs and OCS-Operator throws the error.

`{"level":"error","ts":1679650474.2685232,"logger":"controller-runtime.source","msg":"if kind is a CRD, it should be installed before calling Start","kind":"NooBaa.noobaa.io","error":"no matches for kind \"NooBaa\" in version \"noobaa.io/v1alpha1\"","stacktrace":"sigs.k8s.io/controller-runtime/pkg/source.(*Kind).Start.func1.1\n\t/remote-source/app/vendor/sigs.k8s.io/controller-runtime/pkg/source/source.go:139\nk8s.io/apimachinery/pkg/util/wait.runConditionWithCrashProtectionWithContext\n\t/remote-source/app/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:233\nk8s.io/apimachinery/pkg/util/wait.WaitForWithContext\n\t/remote-source/app/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:660\nk8s.io/apimachinery/pkg/util/wait.poll\n\t/remote-source/app/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:594\nk8s.io/apimachinery/pkg/util/wait.PollImmediateUntilWithContext\n\t/remote-source/app/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:545\nsigs.k8s.io/controller-runtime/pkg/source.(*Kind).Start.func1\n\t/remot...
`

This issue happens even after setting [reconcileStrategy of MCG](https://github.com/red-hat-storage/ocs-operator/blob/release-4.12/api/v1/storagecluster_types.go#L342) to ignore in storageCluster CR

mulbc:
Could we install the CRD and then not have any NooBaa CR?

bindrad:
> Could we install the CRD and then not have any NooBaa CR?

Yes, that is possible. As a workaround, we are going to do the same.

PR #35, opened 2023-03-30
Add initContainer to add CRDs before starting agent controllers

bindrad:
As a temporary workaround, we will create the CRDs on the cluster before controllers are started for the issue mentioned [here](https://github.com/red-hat-storage/managed-fusion-agent/issues/28) and [here](https://github.com/red-hat-storage/managed-fusion-agent/issues/27)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,13 @@ COPY utils/ utils/
 COPY templates/ templates/
 COPY cmd/awsDataGather/ awsDataGather/
 COPY datafoundation/ datafoundation/
+# TODO: Remove these when we have proper solution for CRD problem
+# https://github.com/red-hat-storage/managed-fusion-agent/issues/28
+COPY cmd/crdCreator crdCreator/
+COPY shim/crds/noobaas.noobaa.io.yaml shim/crds/
+COPY shim/crds/ocsinitializations.ocs.openshift.io.yaml shim/crds/
+COPY shim/crds/storageclusters.ocs.openshift.io.yaml shim/crds/
+COPY shim/crds/objectbucketclaims.objectbucket.io.yaml shim/crds/
 
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
@@ -26,13 +33,18 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager 
 # Go will build it as awsDataGather/main; the name will be changed during the copy operation.
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o awsDataGather awsDataGather/main.go
 
+# Build crdCreator binary
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o crdCreator crdCreator/main.go
+
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY --from=builder /workspace/manager .
 COPY --from=builder /workspace/awsDataGather/main awsDataGather
+COPY --from=builder /workspace/crdCreator/main crdCreator
 COPY --from=builder /workspace/templates/customernotification.html /templates/
+COPY --from=builder /workspace/shim/crds/ /shim/crds
 USER nonroot:nonroot
 
 ENTRYPOINT ["/manager"]
--- a/cmd/crdCreator/main.go
+++ b/cmd/crdCreator/main.go
@@ -1,0 +1,67 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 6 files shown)


cmd/crdCreator/main.go
```diff
@@ -0,0 +1,79 @@
...
+//+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;watch;list;create
+
+func main() {
+	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
+	log := ctrl.Log.WithName("main")
+
+	ctx := context.Background()
+
+	log.Info("Setting up dynamic client")
+	dynClient, err := dynamic.NewForConfig(config.GetConfigOrDie())
```
leelavg, line 30: - majorly dynamic client is created using inference of GVK from an unknown bytes
- here the use-case is different, we already know the GVK/GVR of the resource why not use a typed client, convert yaml to json and send the request, something like
``` go
// ref: https://github.com/kubernetes/apimachinery/blob/release-1.27/pkg/util/yaml/decoder.go#L108-L135
import  (
"os"
"k8s.io/apimachinery/pkg/util/yaml"
"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
)
// ..
crd := &v1.CustomResourceDefinition{}
f, _ := os.Open("YAML_FILE")
d := yaml.NewYAMLToJSONDecoder(f)
_ = d.Decode(crd)
// ..
// now you've typed struct, can create typed client-set and perform the op
```

bindrad: Updated

cmd/crdCreator/main.go
```diff
@@ -0,0 +1,79 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+const baseDir string = "shim/crds"
+
+//+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;watch;list;create
+
+func main() {
+	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
```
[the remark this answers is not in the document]
bindrad: I think using `UseDevMode` is fine because it gives a better understanding of errors.

Dockerfile
```diff
@@ -17,6 +17,11 @@ COPY utils/ utils/
 COPY templates/ templates/
 COPY cmd/awsDataGather/ awsDataGather/
 COPY datafoundation/ datafoundation/
+COPY cmd/crdCreator crdCreator/
+COPY shim/crds/noobaas.noobaa.io.yaml shim/crds/
+COPY shim/crds/ocsinitializations.ocs.openshift.io.yaml shim/crds/
+COPY shim/crds/storageclusters.ocs.openshift.io.yaml shim/crds/
+COPY shim/crds/objectbucketclaims.objectbucket.io.yaml shim/crds/
```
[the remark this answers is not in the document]
bindrad: Updated

ezio-auditore (approved):

cmd/crdCreator/main.go
```diff
@@ -0,0 +1,66 @@
...
+
+	for _, f := range crdFiles {
+		file, err := os.Open(fmt.Sprintf("%s/%s", baseDir, f))
+		if err != nil {
+			log.Error(err, "unable to read crd templates")
+			os.Exit(1)
+		}
+		crd := &v1.CustomResourceDefinition{}
+		decoder := yaml.NewYAMLToJSONDecoder(file)
+		_ = decoder.Decode(crd)
```
leelavg, line 57: - any particular reason for not handling error here?

cmd/crdCreator/main.go
```diff
@@ -0,0 +1,66 @@
...
+		if err := k8sClient.Create(ctx, crd); err != nil && !apierrors.IsAlreadyExists(err) {
```
leelavg, line 58: - `Create` on a kubeclient is similar to, get a mapper, init a discovery, infer the traits of this resource and perform this op
- My previous comment is to use typed client, which just performs this op, any reason in not using it?

cmd/crdCreator/main.go
```diff
@@ -0,0 +1,71 @@
...
+	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/yaml"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
```
leelavg, line 18: - I don't think relatively heavy-weight controller-runtime is required here, should get away with client-go itself

cmd/crdCreator/main.go
```diff
@@ -0,0 +1,71 @@
...
+func main() {
+	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
+	log := ctrl.Log.WithName("main")
+
+	ctx := context.Background()
+
+	log.Info("Setting up k8s client")
+	scheme := runtime.NewScheme()
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(apiextensionsv1.AddToScheme(scheme))
```
leelavg, line 34: - I don't think we need a scheme, I'm not totally confident on this though

bindrad: Checked, we don't need it

cmd/crdCreator/main.go
bindrad: Removed them

leelavg (approved):

PR #42, opened 2023-04-05
Fix uninstall flow of DF offering

bindrad:
Issues fixed in the PR

- Missing deletion of storageCluster
- Missing deletion of offering namespace
- Logic was not correct as the reconcile function of offering was not executed once MFO CR was deleted
- Missing add to scheme and rbacs

### DIFF
--- a/controllers/datafoundation.go
+++ b/controllers/datafoundation.go
@@ -66,7 +66,7 @@ type dataFoundationReconciler struct {
 
 //+kubebuilder:rbac:groups=ocs.openshift.io,namespace=system,resources=storageclusters,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=ocs.openshift.io,namespace=system,resources=ocsinitializations,verbs=get;list;watch;update;patch
-//+kubebuilder:rbac:groups=ocs.openshift.io,namespace=system,resources=storageconsumers,verbs=get;list
+//+kubebuilder:rbac:groups=ocs.openshift.io,namespace=system,resources=storageconsumers,verbs=get;list;watch
 //+kubebuilder:rbac:groups=network.openshift.io,resources=ingressnetworkpolicies,verbs=create;get;list;watch;update
 //+kubebuilder:rbac:groups=operators.coreos.com,namespace=system,resources=clusterserviceversions,verbs=get;list;watch;delete;update;patch
 //+kubebuilder:rbac:groups=ocs.openshift.io,namespace=system,resources=ocsinitializations,verbs=get;list;watch;update;patch;delete
@@ -90,6 +90,7 @@ func dfSetupWatches(controllerBuilder *builder.Builder) {
 
 func DFAddToScheme(scheme *runtime.Scheme) {
 	utilruntime.Must(ocsv1.AddToScheme(scheme))
+	utilruntime.Must(ocsv1alpha1.AddToScheme(scheme))
 }
 
 func dfIsReadyToBeRemoved(r *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) (bool, error) {
@@ -150,6 +151,14 @@ func (r *dataFoundationReconciler) initReconciler(reconciler *ManagedFusionOffer
 
 func (r *dataFoundationReconciler) reconcilePhases() (ctrl.Result, error) {
 	if !r.offering.DeletionTimestamp.IsZero() {
+		if err := r.get(&r.storageCluster); err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to get storagecluster: %v", err)
+		}
+		if r.storageCluster.Status.Phase != "Ready" {
+			r.Log.Info("storagecluster is not in ready state, cannot proceed with uninstallation")
+			return ctrl.Result{}, nil
+		}
+
 		storageConsumerList := ocsv1alpha1.StorageConsumerList{}
 		if err := r.list(&storageConsumerList, client.Limit(1)); err != nil {
 			return ctrl.Result{}, fmt.Errorf("unable to list OCS StorageConsumers CRs, %v", err)
@@ -158,6 +167,12 @@ func (r *dataFoundationReconciler) reconcilePhases() (ctrl.Result, error) {
 			r.Log.Info("Found OCS storage consumers, cannot proceed with uninstallation")
 			return ctrl.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
 		}
+
+		r.Log.Info("issuing a delete for storagecluster")
+		if err := r.delete(&r.storageCluster); err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to issue a delete for storagecluster: %v", err)
+		}
+
 	} else {
 		if err := r.reconcileOnboardingValidationSecret(); err != nil {
 			return ctrl.Result{}, err
--- a/controllers/managedfusionoffering_controller.go
+++ b/controllers/managedfusionoffering_controller.go
@@ -26,8 +26,10 @@ import (
 	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	v1alpha1 "github.com/red-hat-storage/managed-fusion-agent/api/v1alpha1"
 	"github.com/red-hat-storage/managed-fusion-agent/utils"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -81,7 +83,12 @@ func (r *ManagedFusionOfferingReconciler) Reconcile(ctx context.Context, req ctr
 	r.initReconciler(ctx, req)
 
 	if err := r.get(r.managedFusionOffering); err != nil {
-		return ctrl.Result{}, fmt.Errorf("unable to get managedFusionOffering: %v", err)
+		if errors.IsNotFound(err) {
+			r.Log.V(-1).Info("ManagedFusionOffering resource not found")
+			return ctrl.Result{}, nil
+		} else {
+			return ctrl.Result{}, fmt.Errorf("unable to get managedFusionOffering: %v", err)
+		}
 	}
 
 	if result, err := r.reconcilePhases(); err != nil {
@@ -129,6 +136,14 @@ func (r *ManagedFusionOfferingReconciler) reconcilePhases() (reconcile.Result, e
 					return ctrl.Result{}, fmt.Errorf("failed to remove finalizer from %s secret: %v", managedFusionSecretName, err)
 				}
 			}
+			r.Log.Info(fmt.Sprintf("issuing a delete for %s namespace", r.namespace))
+			if err := r.delete(&corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: r.namespace,
+				},
+			}); err != nil {
+				return ctrl.Result{}, fmt.Errorf("unable to issue a delete for namespace: %v", err)
+			}
 		}
 	} else if r.managedFusionOffering.UID != "" {
 		if added := utils.AddFinalizer(r.managedFusionOffering, managedFusionFinalizer); added {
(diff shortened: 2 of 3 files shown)


leelavg:
- could you pls add to the commit msg about the issue & bug as this PR is not a feature implementation.

controllers/managedfusionoffering_controller.go
```diff
@@ -81,7 +83,12 @@ func (r *ManagedFusionOfferingReconciler) Reconcile(ctx context.Context, req ctr
 	r.initReconciler(ctx, req)
 
 	if err := r.get(r.managedFusionOffering); err != nil {
-		return ctrl.Result{}, fmt.Errorf("unable to get managedFusionOffering: %v", err)
+		if errors.IsNotFound(err) {
+			r.Log.V(-1).Info("ManagedOCS resource not found")
```
leelavg, line 87:
```suggestion
			r.Log.V(-1).Info("ManagedFusionOffering resource not found")
```

controllers/managedfusionoffering_controller.go
```diff
@@ -147,15 +162,13 @@ func (r *ManagedFusionOfferingReconciler) reconcilePhases() (reconcile.Result, e
...
-			return result, nil
-		}
 	}
 
-	return ctrl.Result{}, nil
+	if result, err := pluginReconcile(r, r.managedFusionOffering); err != nil {
+		return ctrl.Result{}, fmt.Errorf("an error was encountered during reconcile: %v", err)
+	} else {
+		return result, nil
+	}
```
leelavg, line 171:
```suggestion
	result, err := pluginReconcile(r, r.managedFusionOffering)
	if err != nil {
		return ctrl.Result{}, fmt.Errorf("an error was encountered during reconcile: %v", err)
	}
	return result, nil
```

controllers/datafoundation.go
```diff
@@ -158,6 +159,18 @@ func (r *dataFoundationReconciler) reconcilePhases() (ctrl.Result, error) {
 			r.Log.Info("Found OCS storage consumers, cannot proceed with uninstallation")
 			return ctrl.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
 		}
+
+		if err := r.get(&r.storageCluster); err != nil {
```
leelavg, line 163: - I still can't believe this was missed earlier, are you sure that there was no legit reason behind it?

controllers/managedfusionoffering_controller.go
leelavg, line 1: - I'm just thinking it loud, not directly related to this PR, are we intentionally not checking the spec.kind in mfo for setup & deletion?

controllers/datafoundation.go
```diff
@@ -158,6 +159,18 @@ func (r *dataFoundationReconciler) reconcilePhases() (ctrl.Result, error) {
...
+			return ctrl.Result{}, fmt.Errorf("unable to get storagecluster: %v", err)
+		}
+		if r.storageCluster.Status.Phase == "Ready" {
+			r.Log.Info("deleting storagecluster")
+			if err := r.delete(&r.storageCluster); err != nil {
+				return ctrl.Result{}, fmt.Errorf("unable to delete storagecluster: %v", err)
+			}
+		} else {
+			r.Log.Info("storagecluster is not in ready state, cannot proceed with uninstallation")
```
leelavg, line 172: - you need to requeue the event here

controllers/managedfusionoffering_controller.go
```diff
@@ -129,6 +136,14 @@ func (r *ManagedFusionOfferingReconciler) reconcilePhases() (reconcile.Result, e
 					return ctrl.Result{}, fmt.Errorf("failed to remove finalizer from %s secret: %v", managedFusionSecretName, err)
 				}
 			}
+			r.Log.Info(fmt.Sprintf("deleting %s namespace", r.namespace))
+			if err := r.delete(&corev1.Namespace{
```
leelavg, line 140: - I had to think a lot to justify that the flow is correct, could you pls confirm my understanding is correct
1. storagecluster is in ready state & no consumers are attached
2. mfo-cr received a delete event
3. this pr makes sure to actually delete storagecluster cr
4. only after confirming storagecluster cr is absent we'll proceed with ns deletion and implicitly storagecluster cr absence ensures no df things hand around

controllers/managedfusionoffering_controller.go
```diff
@@ -129,6 +136,14 @@ func (r *ManagedFusionOfferingReconciler) reconcilePhases() (reconcile.Result, e
 					return ctrl.Result{}, fmt.Errorf("failed to remove finalizer from %s secret: %v", managedFusionSecretName, err)
 				}
 			}
+			r.Log.Info(fmt.Sprintf("deleting %s namespace", r.namespace))
```
leelavg, line 139: - a couple of lines above, log seems to be incorrect, could you pls replace `secret` with `offering`?

controllers/managedfusionoffering_controller.go
```diff
@@ -129,6 +136,14 @@ func (r *ManagedFusionOfferingReconciler) reconcilePhases() (reconcile.Result, e
...
+			if err := r.delete(&corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: r.namespace,
```
leelavg, line 142: - do we already have rbac in-place to delete ns?

controllers/datafoundation.go
bindrad: I don't think it is required because we are watching storagecluster so if any event occurs on storagecluster CR, there will be a request to the reconciler.

controllers/managedfusionoffering_controller.go
bindrad: Yes, correct

controllers/managedfusionoffering_controller.go
bindrad: Yes, we do. Kubebuilder marker is mentioned in managedFusion Controller.

controllers/managedfusionoffering_controller.go
bindrad: We will check it once we have the code for DFC offering. You can refer to the PR raised for DFC offering.

controllers/datafoundation.go
bindrad: I was not able to think of a reason why it was not present at the first place

controllers/managedfusionoffering_controller.go
bindrad: Updated

controllers/managedfusionoffering_controller.go

controllers/datafoundation.go
leelavg: - I don't think status update results in a reconcile event 🤔 
- Anyways if this is similar to how we do things in deployer, it's fine with me

leelavg (dismissed):

nb-ohad:
@bindrad I think the approach taken here is not optimal, especially since we already have an uninstall-related hook. 
Right now, to accomplish uninstallation for an offering, the offering team needs to implement multiple pieces via multiple hooks, which seems unnecessary to me. 

 I will rethink the uninstallation flow on Sunday and we can discuss it on Monday

controllers/datafoundation.go
bindrad: It should trigger the reconcile event. That's the reason we use generation change predicate where we don't want to reconcile if status/metadata field is changed.

ezio-auditore (approved):

controllers/managedfusionoffering_controller.go
```diff
@@ -147,15 +180,14 @@ func (r *ManagedFusionOfferingReconciler) reconcilePhases() (reconcile.Result, e
...
-		if result, err := pluginReconcile(r, r.managedFusionOffering); err != nil {
-			return ctrl.Result{}, fmt.Errorf("an error was encountered during reconcile: %v", err)
-		} else {
-			return result, nil
-		}
+	if result, err := pluginReconcile(r, r.managedFusionOffering); err != nil {
+		return ctrl.Result{}, fmt.Errorf("an error was encountered during reconcile: %v", err)
+	} else {
+		return result, nil
 	}
```
leelavg, line 189: is there no complaint from any linter that the function is missing final return?

leelavg (approved):

controllers/managedfusionoffering_controller.go
bindrad: No complaints from any linter. We have return in the else so linter does not see it as an issue.

controllers/managedfusionoffering_controller.go
```diff
@@ -81,7 +83,12 @@ func (r *ManagedFusionOfferingReconciler) Reconcile(ctx context.Context, req ctr
 	r.initReconciler(ctx, req)
 
 	if err := r.get(r.managedFusionOffering); err != nil {
-		return ctrl.Result{}, fmt.Errorf("unable to get managedFusionOffering: %v", err)
+		if errors.IsNotFound(err) {
+			r.Log.V(-1).Info("ManagedFusionOffering resource not found")
+			return ctrl.Result{}, nil
```
nb-ohad, line 88: How can it be that we will get into a situation where the reconcile was called but the resource does not exist?

controllers/managedfusionoffering_controller.go
```diff
@@ -124,17 +131,43 @@ func (r *ManagedFusionOfferingReconciler) reconcilePhases() (reconcile.Result, e
...
+					return ctrl.Result{}, fmt.Errorf(
+						"failed to remove finalizer from %s (%s): %v",
+						r.managedFusionOffering.Name,
+						r.managedFusionOffering.Namespace,
+						err,
+					)
 				}
 			}
+			r.Log.Info(fmt.Sprintf("issuing a delete to %s namespace", r.namespace))
+			if err := r.delete(&corev1.Namespace{
```
nb-ohad, line 149: Are we sure it will be ok to delete the ns even before we are sure the offering CR got deleted?

controllers/managedfusionoffering_controller.go
```diff
@@ -124,17 +131,43 @@ func (r *ManagedFusionOfferingReconciler) reconcilePhases() (reconcile.Result, e
...
+				},
+			}); err != nil {
+				return ctrl.Result{}, fmt.Errorf("unable to issue a delete to namespace: %v", err)
+			}
 		}
 	} else if r.managedFusionOffering.UID != "" {
 		if added := utils.AddFinalizer(r.managedFusionOffering, managedFusionFinalizer); added {
-			r.Log.V(-1).Info(fmt.Sprintf("finalizer missing on the %s secret resource, adding...", managedFusionSecretName))
+			r.Log.Info(fmt.Sprintf(
+				"finalizer missing on the %s (%s), adding...",
```
nb-ohad, line 160:
```suggestion
				"finalizer missing on %s (%s), adding...",
```

nb-ohad (changes requested):

controllers/managedfusionoffering_controller.go
bindrad: 1. After installation is successful and ManagedFusionOffering CR exists on the cluster
2. To uninstall, we will issue a delete request on ManagedFusionOffering CR
3. After Offering is ready to be removed, we remove the finalizer from ManagedFusionOffering CR
4. This creates a delete event for ManagedFusionOffering CR and reconcile loop runs, trying to search for ManagedFusionOffering CR but it is not able to get it and throws an error.
5. Because the error is returned another reconcile gets triggered so if the ManagedFusionOffering CR is not found then we should not return error

controllers/managedfusionoffering_controller.go
bindrad: I think it is ok to delete the ns before we are sure that offering CR is deleted because control removal of ManagedFusionOffering CR based on finalizer. And once the finalizer is removed, the assumption is that nothing is blocking the removal of the namespace.

controllers/managedfusionoffering_controller.go

controllers/datafoundation.go
```diff
@@ -158,6 +167,12 @@ func (r *dataFoundationReconciler) reconcilePhases() (ctrl.Result, error) {
 			r.Log.Info("Found OCS storage consumers, cannot proceed with uninstallation")
 			return ctrl.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
 		}
+
+		r.Log.Info("issuing a delete on storagecluster")
+		if err := r.delete(&r.storageCluster); err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to issue a delete on storagecluster: %v", err)
```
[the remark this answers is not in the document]

controllers/datafoundation.go
```diff
@@ -158,6 +167,12 @@ func (r *dataFoundationReconciler) reconcilePhases() (ctrl.Result, error) {
 			r.Log.Info("Found OCS storage consumers, cannot proceed with uninstallation")
 			return ctrl.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
 		}
+
+		r.Log.Info("issuing a delete on storagecluster")
```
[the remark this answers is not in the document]

controllers/managedfusionoffering_controller.go
```diff
@@ -124,17 +131,43 @@ func (r *ManagedFusionOfferingReconciler) reconcilePhases() (reconcile.Result, e
...
-					return ctrl.Result{}, fmt.Errorf("failed to remove finalizer from %s secret: %v", managedFusionSecretName, err)
+					return ctrl.Result{}, fmt.Errorf(
+						"failed to remove finalizer from %s (%s): %v",
+						r.managedFusionOffering.Name,
+						r.managedFusionOffering.Namespace,
+						err,
+					)
 				}
 			}
+			r.Log.Info(fmt.Sprintf("issuing a delete to %s namespace", r.namespace))
```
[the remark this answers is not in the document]

leelavg (approved):

nb-ohad (approved):

PR #48, opened 2023-04-24
Replace deleteAllOf with delete to delete all Offering CRs

bindrad:
DeleteAllOf function does not work as expected and it is broken when a specific namespace is not provided to it. 
Issue link: https://github.com/kubernetes-sigs/controller-runtime/issues/1842


controllers/managedfusion_controller.go
```diff
@@ -303,13 +303,19 @@ func (r *ManagedFusionReconciler) reconcilePhases() (reconcile.Result, error) {
...
 			}
+			for i := range offeringList.Items {
+				offering := offeringList.Items[i]
+				if err := r.delete(&offering); err != nil {
+					return ctrl.Result{}, fmt.Errorf("failed to delete ManagedFusionOffering CR, %v", err)
+				}
+			}
+			r.Log.Info("ManagedFusionOffering CRs delete issued, requeuing the event")
+			return ctrl.Result{RequeueAfter: 10}, nil
+
```
nb-ohad, line 318: Multiple issues here:
1. Why are we checking for managed fusion CRs and then listing again?
2. Even if we were unable to delete one of them, we should delete the rest
3. The error message about a delete failure does not identify the specific CR


I would suggest changing the entire flow in the following manner:
```
offerings := v1alpha1.ManagedFusionOfferingList{}
		if err := r.list(&offerings); err != nil {
			return ctrl.Result{}, fmt.Errorf("failed to list ManagedFusionOffering CRs, %v", err)
		}
		if len(offerings,items) > 0 {
			var found = bool
			var errors []error 
			
			for i := range offeringList.Items {
				offering := &offeringList.Items[i]
				if offering.DeletionTimestamp.IsZero() {
					r.Log.Info(
						"a ManagedFusionOffering CR found %s (%s), issuing delete",
						offering.Name,
						offering.Namespace,
					)

					found = true
					if err := r.delete(offering); err != nil {
						wrapped := fmt.Errorf(
							"failed to delete ManagedFusionOffering CR %s (%s), %v", 
							offering.Name,
							offering.Namespace,
							err,
						)
						error = append(errors, wrapped)
					}
				}
			}

			if len(errors) > 0 {
				return ctrl.Result{}, fmt.Errorf("failed to issue delete for one or more ManagedFusionOffering CRs, %v", errors)
			} else {
				r.Log.Info("ManagedFusionOffering CRs delete issued, requeuing the event")
				return ctrl.Result{RequeueAfter: 10}, nil
			}
```

nb-ohad (changes requested):

controllers/managedfusion_controller.go
bindrad: updated

controllers/managedfusion_controller.go
```diff
@@ -299,14 +299,36 @@ func (r *ManagedFusionReconciler) initReconciler(ctx context.Context, req ctrl.R
...
-			} else {
+		offeringList := v1alpha1.ManagedFusionOfferingList{}
+		if err := r.list(&offeringList); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to list ManagedFusionOffering CRs, %v", err)
+		}
+		if len(offeringList.Items) > 0 {
+			var found bool
+			var errors []error
+			for i := range offeringList.Items {
+				offering := offeringList.Items[i]
```
nb-ohad, line 310: This will create a copy, please use a pointer instead.
```suggestion
				offering := &offeringList.Items[i]
```

nb-ohad (changes requested):

nb-ohad (approved):

PR #53, opened 2023-04-27
Remove endpoint check and update operatorGroup for DFC offering

bindrad:
Remove storageProvider endpoint check
- loadbalancer IP that we get from provider storage cluster does not have a scheme
- parsing host from url return empty string if scheme is not mentioned
- parsing loadbalancer IP does not make sense because we not using a part of it
- instead we are using it as a whole

Update operatorGroup to allnamespace scope for DFC

### DIFF
--- a/controllers/datafoundationclient.go
+++ b/controllers/datafoundationclient.go
@@ -2,7 +2,6 @@ package controllers
 
 import (
 	"fmt"
-	"net/url"
 	"strings"
 
 	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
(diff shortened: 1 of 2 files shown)


controllers/datafoundationclient.go
```diff
@@ -141,19 +140,11 @@ func (r *dataFoundationClientReconciler) reconcileStorageClient() error {
 		return fmt.Errorf("invalid provider endpoint, empty string")
 	}
 
-	providerEndpoint, err := url.ParseRequestURI(r.dataFoundationClientSpec.providerEndpoint)
```
leelavg, line 144: - at a minimum it's better to check the input isn't empty

leelavg (changes requested):

controllers/datafoundationclient.go
bindrad: We are checking it.
https://github.com/red-hat-storage/managed-fusion-agent/pull/53/files#diff-82e77c947fe176519786a3f7087431a7642e152d10391aa74b1bbbf7e21cb406R139-R141

leelavg: My bad.

leelavg (approved):

rohan47:
LGTM

rohan47 (approved):

PR #54, opened 2023-04-27
Restart operator after OCS CSV is installed

rohan47:
Restart operator to register the watchers after ocs operator csv is installed


ezio-auditore (changes requested):

controllers/datafoundationclient.go
```diff
@@ -47,16 +48,27 @@ func dfcSetupWatches(controllerBuilder *builder.Builder) {
...
 }
 
 func DFCAddToScheme(scheme *runtime.Scheme) {
 	utilruntime.Must(ocsclient.AddToScheme(scheme))
 }
 
 func dfcReconcile(offeringReconciler *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) (ctrl.Result, error) {
+	if !DoesCRDExistsAtStart {
+		if utils.DoesCRDsExists("storageclients.ocs.openshift.io") {
+			return ctrl.Result{}, &utils.FatalError{Msg: "need to restart operator"}
```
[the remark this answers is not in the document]
rohan47: changing the msg to `operator needs to restart`

utils/utils.go
```diff
@@ -173,3 +189,31 @@ func HTTPGetAndParseBody(endpoint string) (string, error) {
...
+	}
+
+	crd := apiextensionsv1.CustomResourceDefinition{}
+	err = k8sClient.Get(context.Background(), types.NamespacedName{Name: name}, &crd)
+
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return false
+		} else {
+			setupLog.Error(err, fmt.Sprintf("Error Getting CRD for %s", name))
```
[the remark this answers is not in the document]
rohan47: Done

ezio-auditore (dismissed):

controllers/managedfusionoffering_controller.go
```diff
@@ -45,6 +46,8 @@ const (
 	subscriptionName  = "managed-fusion-offering"
 )
 
+var DoesCRDExistsAtStart bool
```
leelavg, line 49: - I don't see this is being used outside of pkg, can be unexported

utils/utils.go
```diff
@@ -17,19 +17,35 @@ limitations under the License.
...
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
+type FatalError struct {
+	Msg string
+}
+
+func (e *FatalError) Error() string {
+	return e.Msg
+}
```
leelavg, line 47: - maybe a nit, I particularly don't like this for two reasons.
1. type is created here but returned & used in different pkg
2. at call site, this isn't being asserted efficiently (for want of a better word)
- I propose below, pls give a thought about it (I believe it should work)
---
requirement: signal the caller to os exist when a condition is met
sol: use error value in conjunction with `errors.Is`, something like
``` go
package controllers

import (
  "os"
  "errors"
)

const errRestart = "Restart"

// some sample func
err := func() error { return errRestart }
if errors.Is(err, errRestart) { os.Exit(1) }
```

cmd/crdCreator/main.go
```diff
@@ -36,12 +36,9 @@ func main() {
 	}
 
 	crdFiles := []string{
-		"storageclusters.ocs.openshift.io.yaml",
-		"ocsinitializations.ocs.openshift.io.yaml",
 		"noobaas.noobaa.io.yaml",
 		"objectbucketclaims.objectbucket.io.yaml",
 		"objectbuckets.objectbucket.io.yaml",
-		"storageclients.ocs.openshift.io.yaml",
```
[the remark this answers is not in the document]
rohan47: Removed

controllers/managedfusionoffering_controller.go
rohan47: done

utils/utils.go
rohan47: Done

utils/utils.go
```diff
@@ -173,3 +181,31 @@ func HTTPGetAndParseBody(endpoint string) (string, error) {
...
+		if errors.IsNotFound(err) {
+			return false
+		} else {
+			setupLog.Error(err, fmt.Sprintf("Error fetching CRD for %s", name))
+			os.Exit(1)
+		}
+	}
+
+	return true
+}
```
bindrad, line 211: We have a similar function in main.go(`mapCRDAvailability`) so why can't we reuse it? 
Maybe use the util over there as well

controllers/datafoundation.go
```diff
@@ -111,6 +118,13 @@ func dfIsReadyToBeRemoved(r *ManagedFusionOfferingReconciler, offering *v1alpha1
 }
 
 func dfReconcile(offeringReconciler *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) (ctrl.Result, error) {
+	// Check if the required CRDs are installed or not
+	if !doesCRDExistsAtStart {
+		// we only check for storagecluster CRD because other crds are installed as a dependency via ocs-operator
+		if utils.DoesCRDsExists("storageclusters.ocs.openshift.io") {
+			return ctrl.Result{}, errRestart
```
bindrad, line 125: I don't think this is an error. We are restarting because the offering installation succeeded.

controllers/datafoundationclient.go
```diff
@@ -47,16 +48,27 @@ func dfcSetupWatches(controllerBuilder *builder.Builder) {
...
 }
 
 func DFCAddToScheme(scheme *runtime.Scheme) {
 	utilruntime.Must(ocsclient.AddToScheme(scheme))
 }
 
 func dfcReconcile(offeringReconciler *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) (ctrl.Result, error) {
+	if !doesCRDExistsAtStart {
+		if utils.DoesCRDsExists("storageclients.ocs.openshift.io") {
+			return ctrl.Result{}, errRestart
```
bindrad, line 68: I don't think this is an error. We are restarting because the offering installation succeeded.

utils/utils.go
rohan47: What about using this function in `main.go` and updating it to be similar to main.go ?

bindrad: Yeah, that's what I meant. Update the function in utils.go to be similar to main.go function. Then use that util everywhere.

bindrad (dismissed):

controllers/datafoundationclient.go
rohan47: done

leelavg (dismissed):
- exit seems to be better than raising err in this particular case

controllers/datafoundation.go
```diff
@@ -79,13 +81,20 @@ func dfSetupWatches(controllerBuilder *builder.Builder) {
...
-		Owns(&ocsv1.StorageCluster{}).
 		Owns(&netv1.NetworkPolicy{}).
 		Owns(&corev1.ConfigMap{}).
-		Owns(&opv1a1.ClusterServiceVersion{}, csvPredicates).
-		Owns(&ocsv1.OCSInitialization{})
+		Owns(&opv1a1.ClusterServiceVersion{}, csvPredicates)
+
+	// we only check for storagecluster CRD because other crds are installed as a dependency via ocs-operator
+	if utils.MapCRDAvailability(storageClusterCRD)[storageClusterCRD] {
+		doesCRDExistsAtStart = true
```
nb-ohad, line 93: Is that a global variable? If so it is not differentiated between reconcile calls. We should not do that.

controllers/datafoundation.go
```diff
@@ -110,6 +119,14 @@ func dfIsReadyToBeRemoved(r *ManagedFusionOfferingReconciler, offering *v1alpha1
 }
 
 func dfReconcile(offeringReconciler *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) (ctrl.Result, error) {
+	// Check if the required CRDs are installed or not
+	if !doesCRDExistsAtStart {
```
nb-ohad, line 123: Why do we care about existence? what we do care about is if we are watching or not watching specific CRs and we already know that this is based on the CRD map that was loaded in `main.go` and provided to the reconciler.

Also, we should not couple all CRDs under the same umbrella. storagecluster and ocsinitialization are different CRDs and should be treated as such.

controllers/datafoundation.go
```diff
@@ -79,13 +81,20 @@ func dfSetupWatches(controllerBuilder *builder.Builder) {
...
 		Owns(&corev1.Secret{}).
-		Owns(&ocsv1.StorageCluster{}).
 		Owns(&netv1.NetworkPolicy{}).
 		Owns(&corev1.ConfigMap{}).
-		Owns(&opv1a1.ClusterServiceVersion{}, csvPredicates).
-		Owns(&ocsv1.OCSInitialization{})
+		Owns(&opv1a1.ClusterServiceVersion{}, csvPredicates)
+
+	// we only check for storagecluster CRD because other crds are installed as a dependency via ocs-operator
+	if utils.MapCRDAvailability(storageClusterCRD)[storageClusterCRD] {
```
nb-ohad, line 92: Why are we running this here? You can read the same from the CRD map already provided to the reconcile

controllers/datafoundationclient.go
```diff
@@ -47,16 +50,29 @@ func dfcSetupWatches(controllerBuilder *builder.Builder) {
 		),
 	)
 	controllerBuilder.
-		Owns(&opv1a1.ClusterServiceVersion{}, csvPredicates).
-		Owns(&ocsclient.StorageClient{}).
-		Owns(&ocsclient.StorageClassClaim{})
+		Owns(&opv1a1.ClusterServiceVersion{}, csvPredicates)
+
+	if utils.MapCRDAvailability(storageClientCRD)[storageClientCRD] {
```
nb-ohad, line 55: Why are we running this here? You can read the same from the CRD map already provided to the reconcile

controllers/datafoundationclient.go
```diff
@@ -47,16 +50,29 @@ func dfcSetupWatches(controllerBuilder *builder.Builder) {
...
+		doesCRDExistsAtStart = true
```
nb-ohad, line 56: Is that a global variable? If so it is not differentiated between reconcile calls. We should not do that.

controllers/datafoundationclient.go
```diff
@@ -47,16 +50,29 @@ func dfcSetupWatches(controllerBuilder *builder.Builder) {
...
+	}
 }
 
 func DFCAddToScheme(scheme *runtime.Scheme) {
 	utilruntime.Must(ocsclient.AddToScheme(scheme))
 }
 
 func dfcReconcile(offeringReconciler *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) (ctrl.Result, error) {
+	// Check if the required CRDs are installed or not
+	if !doesCRDExistsAtStart {
```
nb-ohad, line 69: Why do we care about existence? what we do care about is if we are watching or not watching specific CRs and we already know that this is based on the CRD map that was loaded in `main.go` and provided to the reconciler.

controllers/datafoundation.go
```diff
@@ -110,6 +119,14 @@ func dfIsReadyToBeRemoved(r *ManagedFusionOfferingReconciler, offering *v1alpha1
 }
 
 func dfReconcile(offeringReconciler *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) (ctrl.Result, error) {
+	// Check if the required CRDs are installed or not
+	if !doesCRDExistsAtStart {
+		// we only check for storagecluster CRD because other crds are installed as a dependency via ocs-operator
+		if utils.MapCRDAvailability(storageClusterCRD)[storageClusterCRD] {
+			offeringReconciler.Log.Info("operator needs to restart to register watchers for offering CRDs")
+			os.Exit(1)
```
nb-ohad, line 127: Don't exist from within an offering code. Return a predefined error that the outer reconcile code could translate into  an `os.Exist`

controllers/datafoundationclient.go
```diff
@@ -47,16 +50,29 @@ func dfcSetupWatches(controllerBuilder *builder.Builder) {
...
+		if utils.MapCRDAvailability(storageClientCRD)[storageClientCRD] {
+			offeringReconciler.Log.Info("operator needs to restart to register watchers for offering CRDs")
+			os.Exit(1)
```
nb-ohad, line 72: Don't exist from within an offering code. Return a predefined error that the outer reconcile code could translate into  an `os.Exist`

controllers/managedfusionoffering_controller.go
```diff
@@ -45,6 +45,8 @@ const (
 	subscriptionName  = "managed-fusion-offering"
 )
 
+var doesCRDExistsAtStart bool
```
nb-ohad, line 48: We should not use global variables as they are not tied to any reconciliation context (as multipole can run in parallel) and on top of that we should not do cross package global variables (the offering code needs to move to the offering package, this is an open tech debt and we should consider that when adding new code)

main.go
```diff
@@ -163,35 +159,6 @@ func getUnrestrictedClient() client.Client {
...
-		if err != nil && !errors.IsNotFound(err) {
-			setupLog.Error(err, fmt.Sprintf("Error Getting CRD for %s", crdName))
-			os.Exit(1)
-		}
-
-		CRDExist[crdName] = crd.UID != ""
-	}
-
-	return CRDExist
-}
```
nb-ohad, line 193: Why are we moving this into the util package? On a general note, we should read all CRDs only once!

nb-ohad (changes requested):

main.go
nb-ohad: That is exactly the point!

rohan47: Then on what basis we should restart the operator? We need to check if OCS related CRDs are present or not, so moved this function in util to update the list of CRDs when required by the offering reconcilers

nb-ohad: Just fetch the current state of the existence of the CRD in question and compare it to the information in the CRD map

nb-ohad: That would be the approach that I will take

rohan47:
Testing out the code changes on a cluster

controllers/managedfusionoffering_controller.go
```diff
@@ -45,11 +47,14 @@ const (
 	subscriptionName  = "managed-fusion-offering"
 )
 
+var errRestart = errors.New("Restart")
```
nb-ohad, line 50: This is not what I was requesting. 
Returning an error for a restart is exactly the same as just `os.Exist` in-place.
The point was to create a new class of errors (preferably in a util file) that forces the operator to exist. 
This does not mean that the error does not have a message or a reason, or even a nested error!

main.go
```diff
@@ -118,24 +118,26 @@ func main() {
 		setupLog.Error(err, "Failed to get 'NAMESPACE' environment variable")
 		os.Exit(1)
 	}
+	availableCRDs := mapCRDAvailability(controllers.EgressFirewallCRD, controllers.EgressNetworkPolicyCRD, controllers.StorageClusterCRD, controllers.StorageClientCRD, controllers.OCSInitializationCRD)
```
nb-ohad, line 121: Why are we passing a list of CRD names here? I believe we agreed that the map will contain an existence mapping for all CRDs installed on the cluster.

main.go
```diff
@@ -163,12 +165,19 @@ func getUnrestrictedClient() client.Client {
...
+	controllers.DFCAddToScheme(scheme)
+}
+
 func mapCRDAvailability(crdNames ...string) map[string]bool {
 
 	var options client.Options
 	options.Scheme = runtime.NewScheme()
 	utilruntime.Must(apiextensionsv1.AddToScheme(options.Scheme))
 
+	setupLog := ctrl.Log.WithName("setup")
```
nb-ohad, line 180: Any reason for this?

main.go
```diff
@@ -163,12 +165,19 @@ func getUnrestrictedClient() client.Client {
 	return k8sClient
 }
 
+// This function is a placeholder for offering plugin integration
+func pluginAddToScheme(scheme *runtime.Scheme) {
+	controllers.DFAddToScheme(scheme)
+	controllers.DFCAddToScheme(scheme)
+}
+
 func mapCRDAvailability(crdNames ...string) map[string]bool {
```
nb-ohad, line 174: As discussed over the last review, no need to pass a CRD name list
```suggestion
func mapCRDAvailability() map[string]bool {
```

utils/utils.go
```diff
@@ -182,3 +189,44 @@ func ToJsonOrDie(value interface{}) []byte {
...
+	if err != nil {
+		setupLog.Error(err, fmt.Sprintf("Error Getting CRD for %s", crdName))
+		return false
+	} else if len(crd.Items) == 0 {
+		return false
+	} else if crd.Items[0].Name == crdName {
+		return true
+	}
+	return false
+}
```
nb-ohad, line 223: This entire function can be replaced with a single `r.Get` on the call sites.  Is there any reason for having it and doing the existence check this way?

nb-ohad (changes requested):

main.go
```diff
@@ -118,24 +117,25 @@ func main() {
 		setupLog.Error(err, "Failed to get 'NAMESPACE' environment variable")
 		os.Exit(1)
 	}
-
+	availableCRDs := mapCRDAvailability()
```
nb-ohad, line 120: Add space as this var is going to be used in multiple sections below,
```suggestion
	availableCRDs := mapCRDAvailability()
	
```

main.go
```diff
@@ -177,23 +184,17 @@ func mapCRDAvailability(crdNames ...string) map[string]bool {
...
-		err = k8sClient.Get(context.Background(), types.NamespacedName{Name: crdName}, &crd)
+	crds := apiextensionsv1.CustomResourceDefinitionList{}
+	err = k8sClient.List(context.Background(), &crds)
 
-		if err != nil && !errors.IsNotFound(err) {
-			setupLog.Error(err, fmt.Sprintf("Error Getting CRD for %s", crdName))
-			os.Exit(1)
-		}
+	if err != nil && !errors.IsNotFound(err) {
+		setupLog.Error(err, fmt.Sprintf("Error Getting CRDs installed on the cluster"))
```
nb-ohad, line 191: The message here is fixed. no need for SprintF 
```suggestion
		setupLog.Error(err, "Error Getting CRDs installed on the cluster")
```

main.go
```diff
@@ -177,23 +184,17 @@ func mapCRDAvailability(crdNames ...string) map[string]bool {
...
+		os.Exit(1)
+	}
 
-		CRDExist[crdName] = crd.UID != ""
+	for _, crd := range crds.Items {
+		CRDExist[crd.Name] = true
```
nb-ohad, line 196: This way of looping over a list will create a copy of each CRD struct on that list as it is passed by value and not by reference. Please use the following instead
```suggestion
	for i := range crds.Items {
		CRDExist[crds.Items[i].Name] = true
```

utils/utils.go
```diff
@@ -31,6 +32,24 @@ import (
...
+
+func HandleOfferingError(err error) error {
+	if err != nil {
+		var errAgentRestart *ErrAgentRestart
+		if errors.As(err, &errAgentRestart) {
+			log.Fatalf("restarting the agent %s", err)
+		}
+	}
+	return err
+}
```
nb-ohad, line 51: This entire error implementation does not make sense.
Please follow go type (and error) extension best practices

controllers/managedfusionoffering_controller.go
```diff
@@ -71,7 +72,7 @@ func (r *ManagedFusionOfferingReconciler) SetupWithManager(mgr ctrl.Manager, ctr
 		Owns(&opv1a1.CatalogSource{}).
 		Owns(&opv1a1.Subscription{})
 
-	pluginSetupWatches(controllerBuilder)
+	pluginSetupWatches(controllerBuilder, r)
```
nb-ohad, line 75:
```suggestion
	pluginSetupWatches(r, controllerBuilder)
```

controllers/datafoundationclient.go
```diff
@@ -52,16 +55,34 @@ func dfcSetupWatches(controllerBuilder *builder.Builder) {
...
 
 func dfcReconcile(offeringReconciler *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) (ctrl.Result, error) {
+
+	if !offeringReconciler.AvailableCRDs[StorageClientCRD] {
+		// Check if the required CRDs are installed or not
+		crd := ocsclient.StorageClient{}
+		err := offeringReconciler.get(&crd)
+		if err != nil && !errors.IsNotFound(err) {
+			return ctrl.Result{}, &utils.ErrAgentRestart{Msg: "restart agent to register watchers for the CRDs"}
+		}
```
nb-ohad, line 83: The logic seems broken here. You are restarting if there is a get error that is not a NotFound error.
The correct logic is: Get the CRD, if the get operation results in any kind of error then request via error. 
If the Get operation was successful then restart

controllers/datafoundation.go
```diff
@@ -110,6 +121,16 @@ func dfIsReadyToBeRemoved(r *ManagedFusionOfferingReconciler, offering *v1alpha1
...
 
 func dfReconcile(offeringReconciler *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) (ctrl.Result, error) {
+
+	if !offeringReconciler.AvailableCRDs[StorageClusterCRD] {
+		// we only check for storagecluster CRD because other crds are installed as a dependency via ocs-operator
+		crd := ocsv1.StorageCluster{}
+		err := offeringReconciler.get(&crd)
+		if err != nil && !errors.IsNotFound(err) {
+			return ctrl.Result{}, &utils.ErrAgentRestart{Msg: "restart agent to register watchers for the CRDs"}
+		}
```
nb-ohad, line 131: The logic seems broken here. You are restarting if there is a get error that is not a NotFound error.
The correct logic is: Get the CRD, if the get operation results in any kind of error then request via error. 
If the Get operation was successful then restart

controllers/datafoundation.go
```diff
@@ -71,21 +73,30 @@ type dataFoundationReconciler struct {
 //+kubebuilder:rbac:groups=network.openshift.io,resources=ingressnetworkpolicies,verbs=create;get;list;watch;update
 //+kubebuilder:rbac:groups=operators.coreos.com,resources=clusterserviceversions,verbs=get;list;watch;delete;update;patch
 
-func dfSetupWatches(controllerBuilder *builder.Builder) {
+func dfSetupWatches(controllerBuilder *builder.Builder, offeringReconciler *ManagedFusionOfferingReconciler) {
```
nb-ohad, line 76: It is common to call the governing reconciler `r` and have it as the first parameter (to follow all the other 
```suggestion
func dfcSetupWatches(r *ManagedFusionOfferingReconciler, controllerBuilder *builder.Builder) {
```

nb-ohad (changes requested):

utils/utils.go
leelavg: - I believe we (reviewers) are offering different views to how this should be handled
- Author implemented error type but matched it wrongly at first, I told to use error value, Dhruv got rid of error altogether and you have mentioned to use error type
- However, latest comment mentions entire thing is wrong
- In this particular instance, could you pls expand on what you anticipate @nb-ohad ? (in a sense over the review period the requirement/assumptions seems to have changed as well)

below is what I'm aware wrt errors (in current context)
1. there's no need to wrap error here (`%w`)
2. error type & value is checked using `errors.As` and `errors.Is` resp. in general
3. here there's no requirement for a new error (/custom) type, let alone being in a util

I'm just sharing my opinion, happy to be corrected.

nb-ohad: Thank you @leelavg for noticing the different viewpoints of this specific comment. 
My original comment was to delegate a response to this class of errors - one which needs a different response than just a re-queue - to the scope outside of the offering specific code. 

The reason is that it makes sense that the response we have today (restart the operator) will change to something else (maybe, add a dynamic watch). 

My only issue with the current implementation is that it does not follow golang's best practices on type (and error) extensions.  

Saying that I do see @bindrad point that this mechanism is overcomplicated. May I suggest a shift in direction?
Maybe we want a function, somewhere common, that will handle missing watches.  The implementation for that function, for now, will be to `os.Exit` but might change as we figure out a better way. That way the handling becomes very simple, just call the `handleMissingWatch` function when a watch is missing

WDYT?

leelavg: > one which needs a different response than just a re-queue - to the scope outside of the offering specific code.
- atm, we just need a re-queue right

> maybe, add a dynamic watch
- I believe, it's going to be entirely different implementation (based on what I understand from Dhruv's findings)

> golang's best practices on type (and error) extensions.
- I maybe dragging it further, however I'm genuinely interested what's missing here.

> might change as we figure out a better way.
- atm, I can't think of better way to restart operator within itself
- anyways, the way we are doing things here doesn't restrict any new methods that'll be found later on

> That way the handling becomes very simple, just call the handleMissingWatch function when a watch is missing
- sry, I don't understand this, maybe you are looking from a different angle that we are missing, our (at-least my) understanding is, if a specific CRD is available we need to setup watch on it or else do nothing

nb-ohad: @leelavg 
> atm, we just need a re-queue right

No, we need a restart, that is the entire point

> I believe, it's going to be entirely different implementation (based on what I understand from Dhruv's findings)

If we figure out ho to add a watch mid reconcile. I don't believe the flow will be that different

> I maybe dragging it further, however I'm genuinely interested what's missing here.

I will try to find a reference when I have some spare time

> atm, I can't think of better way to restart operator within itself

The restart will be using `os.Exist` the issue here is who own that operation and the relationship between offering code and general control code

> anyways, the way we are doing things here doesn't restrict any new methods that'll be found later on

That is true, but in that case, lets keep it simple and implement my latest suggestion (a function call)

leelavg: > implement my latest suggestion (a function call)
- ack, so a function call which just does `os.Exit` without errors and all

controllers/datafoundation.go
```diff
@@ -243,6 +258,20 @@ func (r *dataFoundationReconciler) parseSpec(offering *v1alpha1.ManagedFusionOff
 	return nil
 }
 
+func (r *dataFoundationReconciler) reconcileWatchers() error {
```
nb-ohad, line 261: Why is this function named `reconcile`? It does not do any kind of reconciliation.  Also, why do we need the function at all? this code should be embedded at the start of reconcilePhases.

controllers/datafoundation.go
```diff
@@ -243,6 +258,20 @@ func (r *dataFoundationReconciler) parseSpec(offering *v1alpha1.ManagedFusionOff
...
+	if !r.ManagedFusionOfferingReconciler.AvailableCRDs[StorageClusterCRD] {
+		crd := apiextensionsv1.CustomResourceDefinition{}
+		crd.Name = StorageClusterCRD
+		// we only check for storagecluster CRD because other crds are installed as a dependency via ocs-operator
```
nb-ohad, line 265: Please don't make such assumptions.

controllers/datafoundation.go
```diff
@@ -171,6 +183,9 @@ func (r *dataFoundationReconciler) reconcilePhases() (ctrl.Result, error) {
 		}
 
 	} else {
+		if err := r.reconcileWatchers(); err != nil {
+			return ctrl.Result{}, err
+		}
```
nb-ohad, line 188: The code in this func needs to run before anything else in the reconcilePhases func as it is an assumption that we need to ensure before having any reconcile code.

controllers/datafoundation.go
```diff
@@ -243,6 +258,20 @@ func (r *dataFoundationReconciler) parseSpec(offering *v1alpha1.ManagedFusionOff
...
 
+func (r *dataFoundationReconciler) reconcileWatchers() error {
+	if !r.ManagedFusionOfferingReconciler.AvailableCRDs[StorageClusterCRD] {
+		crd := apiextensionsv1.CustomResourceDefinition{}
+		crd.Name = StorageClusterCRD
+		// we only check for storagecluster CRD because other crds are installed as a dependency via ocs-operator
+		if err := r.get(&crd); err != nil {
+			return err
+		} else {
+			utils.HandleMissingWatchers()
```
nb-ohad, line 269: How does the util function know which watch is missing? you are not passing any kind of information to the util func

controllers/datafoundationclient.go
```diff
@@ -36,13 +37,15 @@ type dataFoundationClientReconciler struct {
 const (
 	ocsClientOperatorName                = "ocs-client-operator"
 	csiKMSConnectionDetailsConfigMapName = "csi-kms-connection-details"
+	StorageClientCRD                     = "storageclients.ocs.openshift.io"
+	StorageClassClaimCRD                 = "storageclassclaims.ocs.openshift.io"
```
nb-ohad, line 41:
```suggestion
	StorageClientCRDName                     = "storageclients.ocs.openshift.io"
	StorageClassClaimCRDName                 = "storageclassclaims.ocs.openshift.io"
```

controllers/datafoundationclient.go
```diff
@@ -144,6 +158,20 @@ func (r *dataFoundationClientReconciler) reconcilePhases() (ctrl.Result, error)
 	}
 }
 
+func (r *dataFoundationClientReconciler) reconcileWatchers() error {
```
nb-ohad, line 161: Why is this function named reconcile? It does not do any kind of reconciliation. Also, why do we need the function at all? this code should be embedded at the start of reconcilePhases.

controllers/datafoundationclient.go
```diff
@@ -125,6 +136,9 @@ func (r *dataFoundationClientReconciler) reconcilePhases() (ctrl.Result, error)
 	if !r.offering.DeletionTimestamp.IsZero() {
 		return ctrl.Result{}, nil
 	} else {
+		if err := r.reconcileWatchers(); err != nil {
+			return ctrl.Result{}, err
+		}
```
nb-ohad, line 141: The code in this func needs to run before anything else in the reconcilePhases func as it is an assumption that we need to ensure before having any reconcile code.

nb-ohad (changes requested):

utils/utils.go
```diff
@@ -31,6 +31,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
+func HandleMissingWatchers() {
+	log.Fatal("restarting the agent to register watchers")
+}
```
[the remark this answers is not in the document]
rohan47: 3. `log.Fatal` throws a message and exits the overator via `os.Exit(1)`

controllers/datafoundation.go
```diff
@@ -171,6 +181,21 @@ func (r *dataFoundationReconciler) reconcilePhases() (ctrl.Result, error) {
...
+		if !r.ManagedFusionOfferingReconciler.AvailableCRDs[StorageClusterCRDName] && len(missingCRDs) < 1 {
+			missingCRDs = append(missingCRDs, StorageClusterCRDName)
+		}
+		if !r.ManagedFusionOfferingReconciler.AvailableCRDs[OCSInitializationCRDName] && len(missingCRDs) < 2 {
+			missingCRDs = append(missingCRDs, OCSInitializationCRDName)
+		}
+		if len(missingCRDs) != 0 {
+			if err := utils.HandleMissingWatchers(r.Client, missingCRDs); err != nil {
+				return ctrl.Result{}, err
+			}
```
nb-ohad, line 196: @rohan47 This seems like a regression of the older revision. Where are the get operations? why are you keeping a list? why do we have only a single call to `HandleMissingWatchers`?

I would expect to see a simple get+HandleMissingWatchers call, for each of the CRD in question. The simplest solution possible.

controllers/datafoundationclient.go
```diff
@@ -125,6 +135,21 @@ func (r *dataFoundationClientReconciler) reconcilePhases() (ctrl.Result, error)
...
+		if !r.ManagedFusionOfferingReconciler.AvailableCRDs[StorageClassClaimCRDName] && len(missingCRDs) < 2 {
+			missingCRDs = append(missingCRDs, StorageClassClaimCRDName)
+		}
+		if len(missingCRDs) != 0 {
+			if err := utils.HandleMissingWatchers(r.Client, missingCRDs); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
+
```
nb-ohad, line 152: Same as last comment.

utils/handlers.go
```diff
@@ -0,0 +1,44 @@
...
+			return err
+		} else {
+			crdsPresent++
+			log.Printf("agent restart required to register watchers for the CRD %s", crdName[i])
+		}
+	}
+	if crdsPresent == len(crdName) {
+		log.Fatal("all missing CRDs present on cluster, restarting the agent to register watchers")
+	}
+	return nil
```
nb-ohad, line 43: Why does "handle" utils do "get"s?
I would expect it to just be a log operation followed by a fatal operation

Also, handlers as a filename have no context. Please move to the appropriate util file.

nb-ohad (changes requested):

utils/handlers.go
rohan47: renamed the file to `restart.go` this can be renamed or moved to a different file later in the future when there is a suitable util file

controllers/datafoundation.go
```diff
@@ -171,6 +182,27 @@ func (r *dataFoundationReconciler) reconcilePhases() (ctrl.Result, error) {
 		}
 
 	} else {
+
+		// Checking for CRDs that are installed after the agent was started and add watchers for the new CRDs
+		if !r.ManagedFusionOfferingReconciler.AvailableCRDs[StorageClusterCRDName] {
```
nb-ohad, line 187: Please do not access the underlaying reconciler in the offering code.
You can copy a ref to the AvailableCRDs map into the sub reconciler on init

controllers/datafoundation.go
```diff
@@ -42,9 +43,11 @@ import (
...
 
 const (
-	defaultDeviceSetName = "default"
-	ocsOperatorName      = "ocs-operator"
-	storageClusterName   = "ocs-storagecluster"
+	defaultDeviceSetName     = "default"
+	ocsOperatorName          = "ocs-operator"
+	storageClusterName       = "ocs-storagecluster"
+	StorageClusterCRDName    = "storageclusters.ocs.openshift.io"
+	OCSInitializationCRDName = "ocsinitializations.ocs.openshift.io"
```
nb-ohad, line 50: Why do we need these as public constants?

controllers/datafoundationclient.go
```diff
@@ -36,13 +37,15 @@ type dataFoundationClientReconciler struct {
 const (
 	ocsClientOperatorName                = "ocs-client-operator"
 	csiKMSConnectionDetailsConfigMapName = "csi-kms-connection-details"
+	StorageClientCRDName                 = "storageclients.ocs.openshift.io"
+	StorageClassClaimCRDName             = "storageclassclaims.ocs.openshift.io"
```
nb-ohad, line 41: Why do we need these as public constants?

controllers/datafoundation.go
```diff
@@ -171,6 +182,27 @@ func (r *dataFoundationReconciler) reconcilePhases() (ctrl.Result, error) {
...
+		}
+		if !r.ManagedFusionOfferingReconciler.AvailableCRDs[OCSInitializationCRDName] {
+			crd := apiextensionsv1.CustomResourceDefinition{}
+			crd.Name = OCSInitializationCRDName
+			if err := r.get(&crd); err != nil {
+				return ctrl.Result{}, err
+			} else {
+				utils.HandleMissingWatchers(OCSInitializationCRDName)
+			}
+		}
```
nb-ohad, line 204: @rohan47 
 Requesting again for the 3rd time (3rd review) - Move this code to the start of `reconcilePhases`, This code is a prereq to running any of the offering reconcile logic!

controllers/datafoundationclient.go
```diff
@@ -125,6 +136,27 @@ func (r *dataFoundationClientReconciler) reconcilePhases() (ctrl.Result, error)
...
+		}
+		if !r.ManagedFusionOfferingReconciler.AvailableCRDs[StorageClassClaimCRDName] {
+			crd := apiextensionsv1.CustomResourceDefinition{}
+			crd.Name = StorageClassClaimCRDName
+			if err := r.get(&crd); err != nil {
+				return ctrl.Result{}, err
+			} else {
+				utils.HandleMissingWatchers(StorageClassClaimCRDName)
+			}
+		}
```
nb-ohad, line 158: @rohan47
Requesting again for the 3rd time (3rd review) - Move this code to the start of reconcilePhases, This code is a prereq to running any of the offering reconcile logic!

controllers/datafoundation.go
```diff
@@ -171,6 +182,27 @@ func (r *dataFoundationReconciler) reconcilePhases() (ctrl.Result, error) {
...
+		if !r.ManagedFusionOfferingReconciler.AvailableCRDs[StorageClusterCRDName] {
+			crd := apiextensionsv1.CustomResourceDefinition{}
+			crd.Name = StorageClusterCRDName
+			if err := r.get(&crd); err != nil {
+				return ctrl.Result{}, err
+			} else {
+				utils.HandleMissingWatchers(StorageClusterCRDName)
+			}
+		}
+		if !r.ManagedFusionOfferingReconciler.AvailableCRDs[OCSInitializationCRDName] {
```
nb-ohad, line 196: Please do not access the underlaying reconciler in the offering code.
You can copy a ref to the AvailableCRDs map into the sub reconciler on init

controllers/datafoundationclient.go
```diff
@@ -125,6 +136,27 @@ func (r *dataFoundationClientReconciler) reconcilePhases() (ctrl.Result, error)
 	if !r.offering.DeletionTimestamp.IsZero() {
 		return ctrl.Result{}, nil
 	} else {
+
+		// Checking for CRDs that are installed after the agent was started and add watchers for the new CRDs
+		if !r.ManagedFusionOfferingReconciler.AvailableCRDs[StorageClientCRDName] {
```
nb-ohad, line 141: Please do not access the underlaying reconciler in the offering code.
You can copy a ref to the AvailableCRDs map into the sub reconciler on init

controllers/datafoundationclient.go
```diff
@@ -125,6 +136,27 @@ func (r *dataFoundationClientReconciler) reconcilePhases() (ctrl.Result, error)
...
+			crd := apiextensionsv1.CustomResourceDefinition{}
+			crd.Name = StorageClientCRDName
+			if err := r.get(&crd); err != nil {
+				return ctrl.Result{}, err
+			} else {
+				utils.HandleMissingWatchers(StorageClientCRDName)
+			}
+		}
+		if !r.ManagedFusionOfferingReconciler.AvailableCRDs[StorageClassClaimCRDName] {
```
nb-ohad, line 150: Please do not access the underlaying reconciler in the offering code.
You can copy a ref to the AvailableCRDs map into the sub reconciler on init

utils/restart.go
```diff
@@ -0,0 +1,26 @@
...
+limitations under the License.
+*/
+
+package utils
+
+import (
+	"log"
+)
+
+func HandleMissingWatchers(crdName string) {
```
nb-ohad, line 23: Please rename to: `HandleMissingWatch(reconciler, missingCRDName)`

utils/restart.go
```diff
@@ -0,0 +1,26 @@
...
+	log.Printf("agent restart required to register watchers for the CRD %s", crdName)
+	log.Fatal("restarting the agent to register watchers")
```
nb-ohad, line 25:
```suggestion
	log.Fatal("A watch is missing for resource type %s, restarting agent", missingCRDName)
```

utils/restart.go
nb-ohad, line 1: The file name should be describing an area of functionality. `restart` is a bad name. 
As the functions here are expected to be common utils for reconcilers functionality and should accept a reconciler as the first param I would suggest to rename this to: `reconcile`

nb-ohad (changes requested):

utils/restart.go
rohan47: Is `reconciler` really required here? we are not going to use anything from it right?

controllers/datafoundation.go
```diff
@@ -144,9 +156,31 @@ func (r *dataFoundationReconciler) initReconciler(reconciler *ManagedFusionOffer
 
 	r.ocsInitialization.Name = "ocsinit"
 	r.ocsInitialization.Namespace = offering.Namespace
+
+	r.availableCRDs = reconciler.AvailableCRDs
 }
 
 func (r *dataFoundationReconciler) reconcilePhases() (ctrl.Result, error) {
+	// Checking for CRDs that are installed after the agent was started and add watchers for the new CRDs
```
nb-ohad, line 164: We are not adding any watches right now.
```suggestion
	// Checking for CRDs that were installed after the agent was started
```

controllers/datafoundation.go
```diff
@@ -171,6 +205,7 @@ func (r *dataFoundationReconciler) reconcilePhases() (ctrl.Result, error) {
 		}
 
 	} else {
+
```
nb-ohad, line 208:
```suggestion
```

controllers/datafoundationclient.go
```diff
@@ -122,9 +136,30 @@ func (r *dataFoundationClientReconciler) parseSpec(offering *v1alpha1.ManagedFus
 }
 
 func (r *dataFoundationClientReconciler) reconcilePhases() (ctrl.Result, error) {
+	// Checking for CRDs that are installed after the agent was started and add watchers for the new CRDs
```
nb-ohad, line 139: We are not adding any watches right now.
```suggestion
	// Checking for CRDs that were installed after the agent was started
```

nb-ohad (changes requested):

controllers/datafoundationclient.go
rohan47: done

PR #55, opened 2023-05-04
DFC uninstall flow

ezio-auditore:
- checks for PVs using ocs storage classes and restricts delete if found
- delete default storageclass claims
- delete storageclient

### DIFF
--- a/controllers/datafoundationclient.go
+++ b/controllers/datafoundationclient.go
@@ -3,12 +3,15 @@ package controllers
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	v1alpha1 "github.com/red-hat-storage/managed-fusion-agent/api/v1alpha1"
 	"github.com/red-hat-storage/managed-fusion-agent/utils"
 	ocsclient "github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -34,13 +37,20 @@ type dataFoundationClientReconciler struct {
 }
 
 const (
-	ocsClientOperatorName                = "ocs-client-operator"
 	csiKMSConnectionDetailsConfigMapName = "csi-kms-connection-details"
+	ocsClientOperatorName                = "ocs-client-operator"
+	defaultBlockStorageClassClaimName    = "ocs-storagecluster-ceph-rbd"
+	defaultFSStorageClassClaimName       = "ocs-storagecluster-cephfs"
+	defaultStorageClientName             = "ocs-storageclient"
+	ocsCephFsProvisionerSuffix           = ".cephfs.csi.ceph.com"
+	ocsRBDProvisionerSuffix              = ".rbd.csi.ceph.com"
 )
 
 //+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclients,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclassclaims,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=operators.coreos.com,resources=clusterserviceversions,verbs=get;list;watch;delete;update;patch
+//+kubebuilder:rbac:groups="",resources=persistentvolumes,verbs=get;list;watch
+//+kubebuilder:rbac:groups="storage.k8s.io",resources=storageclass,verbs=get;list;watch
 
 func dfcSetupWatches(controllerBuilder *builder.Builder) {
 	csvPredicates := builder.WithPredicates(
@@ -75,12 +85,12 @@ func (r *dataFoundationClientReconciler) initReconciler(reconciler *ManagedFusio
 	r.ManagedFusionOfferingReconciler = reconciler
 	r.offering = offering
 
-	r.storageClient.Name = "ocs-storageclient"
+	r.storageClient.Name = defaultStorageClientName
 	r.storageClient.Namespace = offering.Namespace
 
-	r.defaultBlockStorageClassClaim.Name = "ocs-storagecluster-ceph-rbd"
+	r.defaultBlockStorageClassClaim.Name = defaultBlockStorageClassClaimName
 
-	r.defaultFSStorageClassClaim.Name = "ocs-storagecluster-cephfs"
+	r.defaultFSStorageClassClaim.Name = defaultFSStorageClassClaimName
 
 	r.csiKMSConnectionDetailsConfigMap.Name = csiKMSConnectionDetailsConfigMapName
 	r.csiKMSConnectionDetailsConfigMap.Namespace = offering.Namespace
@@ -123,6 +133,23 @@ func (r *dataFoundationClientReconciler) parseSpec(offering *v1alpha1.ManagedFus
 
 func (r *dataFoundationClientReconciler) reconcilePhases() (ctrl.Result, error) {
 	if !r.offering.DeletionTimestamp.IsZero() {
+		if found, err := r.hasOCSVolumes(); err != nil {
+			return ctrl.Result{}, err
+		} else if found {
+			r.Log.Info("PVs provisioned by ODF  storage classes exists, cannot proceed with uninstallation")
+			return ctrl.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
+		}
+		r.Log.Info("issuing a delete for Default StorageClassClaims")
+		if err := r.delete(&r.defaultBlockStorageClassClaim); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to issue a delete for defaultBlockStorageClassClaim: %v", err)
+		}
+		if err := r.delete(&r.defaultFSStorageClassClaim); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to issue a delete for defaultFSStorageClassClaim: %v", err)
+		}
+		r.Log.Info("issuing a delete for StorageClient")
+		if err := r.delete(&r.storageClient); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to issue a delete for Storage Client: %v", err)
+		}
 		return ctrl.Result{}, nil
 	} else {
 		if err := r.reconcileCSV(); err != nil {
@@ -247,4 +274,53 @@ func (r *dataFoundationClientReconciler) reconcileCSIKMSConnectionDetailsConfigM
 	}
 
 	return nil
+}
+
+func (r *dataFoundationClientReconciler) hasOCSVolumes() (bool, error) {
+	storageClassList := &storagev1.StorageClassList{}
+	if err := r.list(storageClassList); err != nil {
+		return false, fmt.Errorf("failed to list Storage Classes: %v", err)
+	}
+	ocsStorageClass := map[string]bool{}
+	rbdProvisionerName := r.offering.Namespace + ocsRBDProvisionerSuffix
+	fsProvisionerName := r.offering.Namespace + ocsCephFsProvisionerSuffix
+	for i := range storageClassList.Items {
+		storageClass := storageClassList.Items[i]
+		if storageClass.Provisioner == fsProvisionerName || storageClass.Provisioner == rbdProvisionerName {
+			ocsStorageClass[storageClass.Name] = true
+		}
+	}
+	pvList := &corev1.PersistentVolumeList{}
+	pvCount := 0
+	if err := r.UnrestrictedClient.List(r.ctx, pvList); err != nil {
+		return false, fmt.Errorf("failed to list persistent volumes: %v", err)
+	}
+	for i := range pvList.Items {
+		scName := pvList.Items[i].Spec.StorageClassName
+		if ocsStorageClass[scName] {
+			pvCount++
+		}
+	}
+	if pvCount > 0 {
+		r.Log.Info(fmt.Sprintf("Current count of PVs blocking uninstallation: %v", pvCount))
+		return true, nil
+	}
+
+	return false, nil
+}
+
+func dfcIsReadyToBeRemoved(r *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) (bool, error) {
+	storageClient := &ocsclient.StorageClient{}
+	storageClient.Name = defaultStorageClientName
+	storageClient.Namespace = offering.Namespace
+
+	if err := r.get(storageClient); err != nil {
+		if errors.IsNotFound(err) {
+			return true, nil
+		} else {
+			return false, err
+		}
+	} else {
+		return false, nil
+	}
 }
--- a/main.go
+++ b/main.go
@@ -133,9 +133,10 @@ func main() {
 	}
 
 	if err = (&controllers.ManagedFusionOfferingReconciler{
-		Client: mgr.GetClient(),
-		Log:    ctrl.Log.WithName("controllers").WithName("ManagedFusionOffering"),
-		Scheme: mgr.GetScheme(),
+		Client:             mgr.GetClient(),
+		UnrestrictedClient: getUnrestrictedClient(),
+		Log:                ctrl.Log.WithName("controllers").WithName("ManagedFusionOffering"),
+		Scheme:             mgr.GetScheme(),
 	}).SetupWithManager(mgr, nil); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "ManagedFusionOffering")
 		os.Exit(1)
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -27,6 +27,14 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - persistentvolumes
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
   - secrets
   - secrets/finalizers
   verbs:
@@ -267,6 +275,14 @@ rules:
   - list
   - patch
   - update
+  - watch
+- apiGroups:
+  - storage.k8s.io
+  resources:
+  - storageclass
+  verbs:
+  - get
+  - list
   - watch
 ---
 apiVersion: rbac.authorization.k8s.io/v1
--- a/controllers/managedfusionoffering_controller.go
+++ b/controllers/managedfusionoffering_controller.go
@@ -47,9 +47,10 @@ const (
 
 // ManagedFusionOfferingReconciler reconciles a ManagedFusionOffering object
 type ManagedFusionOfferingReconciler struct {
-	Client client.Client
-	Log    logr.Logger
-	Scheme *runtime.Scheme
+	Client             client.Client
+	UnrestrictedClient client.Client
+	Log                logr.Logger
+	Scheme             *runtime.Scheme
 
 	ctx                   context.Context
 	namespace             string
@@ -315,7 +316,13 @@ func (r *ManagedFusionOfferingReconciler) getCSVByPrefix(name string) (*opv1a1.C
 // All the below functions are placeholder for offering plugin integration
 
 func pluginIsReadyToBeRemoved(reconciler *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering) (bool, error) {
-	return dfIsReadyToBeRemoved(reconciler, offering)
+	switch offering.Spec.Kind {
+	case v1alpha1.KindDataFoundation:
+		return dfIsReadyToBeRemoved(reconciler, offering)
+	case v1alpha1.KindDataFoundationClient:
+		return dfcIsReadyToBeRemoved(reconciler, offering)
+	}
+	return false, nil
 }
 
 // This function is a placeholder for offering plugin integration


controllers/datafoundationclient.go
```diff
@@ -73,9 +80,9 @@ func (r *dataFoundationClientReconciler) initReconciler(reconciler *ManagedFusio
 	r.storageClient.Name = "ocs-storageclient"
 	r.storageClient.Namespace = offering.Namespace
 
-	r.defaultBlockStorageClassClaim.Name = "ocs-storagecluster-ceph-rbd"
+	r.defaultBlockStorageClassClaim.Name = defaultBlockStorageClassClaimName
 
-	r.defaultFSStorageClassClaim.Name = "ocs-storagecluster-cephfs"
+	r.defaultFSStorageClassClaim.Name = defaultFSStorageClassClaimName
```
rewantsoni, line 85: @ezio-auditore Are we changing the claimNames? It might create issues during migration as on 4.11 we have claims created with ocs-storagecluster-ceph-rbd and ocs-storagecluster-cephfs

ezio-auditore: Not changing the claim names but refactoring it and having the value as a const

controllers/datafoundationclient.go
```diff
@@ -208,3 +234,28 @@ func (r *dataFoundationClientReconciler) reconcileDefaultFSStorageClassClaim() e
...
+	if err := r.list(&storageClassList); err != nil {
+		return false, fmt.Errorf("failed to list Storage Classes: %v", err)
+	}
+	ocsStorageClass := make(map[string]bool)
+	for i := range storageClassList.Items {
+		storageClassName := &storageClassList.Items[i].Name
+		if *storageClassName == defaultBlockStorageClassClaimName || *storageClassName == defaultFSStorageClassClaimName {
+			ocsStorageClass[*storageClassName] = true
+		}
+	}
```
leelavg, line 249: - declaring the var as pointer seems better considering the rest of the operations 🤔 
```suggestion
	storageClassList := &storagev1.StorageClassList{}
	if err := r.list(storageClassList); err != nil {
		return false, fmt.Errorf("failed to list Storage Classes: %v", err)
	}
	ocsStorageClass := make(map[string]bool)
	for i := range storageClassList.Items {
		storageClassName := storageClassList.Items[i].Name
		if storageClassName == defaultBlockStorageClassClaimName || storageClassName == defaultFSStorageClassClaimName {
			ocsStorageClass[storageClassName] = true
		}
	}
```

leelavg (dismissed):

bindrad (changes requested):
@ezio-auditore I don't see a function that will indicate whether the offering is safe to be removed or not. We should have `dfcIsReadyToBeRemoved` function which will be called in `pluginIsReadyToBeRemoved` function.

leelavg:
> I don't see a function that will indicate
- good call, seems I need to look at agent code before reviewing next PRs 😅

controllers/datafoundationclient.go
```diff
@@ -31,12 +35,17 @@ type dataFoundationClientReconciler struct {
...
+	defaultBlockStorageClassClaimName = "ocs-storagecluster-ceph-rbd"
+	defaultFSStorageClassClaimName    = "ocs-storagecluster-cephfs"
+	defaultStorageClientName          = "ocs-storageclient"
 )
 
 //+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclients,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclassclaims,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=operators.coreos.com,namespace=system,resources=clusterserviceversions,verbs=get;list;watch;delete;update;patch
+// +kubebuilder:rbac:groups="",resources=persistentvolumes,verbs=get;list;watch
+// +kubebuilder:rbac:groups="storage.k8s.io",resources=storageclass,verbs=get;list;watch
```
bindrad, line 48:
```suggestion
//+kubebuilder:rbac:groups="",resources=persistentvolumes,verbs=get;list;watch
//+kubebuilder:rbac:groups="storage.k8s.io",resources=storageclass,verbs=get;list;watch
```

controllers/datafoundationclient.go
```diff
@@ -115,6 +124,25 @@ func (r *dataFoundationClientReconciler) parseSpec(offering *v1alpha1.ManagedFus
...
 func (r *dataFoundationClientReconciler) reconcilePhases() (ctrl.Result, error) {
 	if !r.offering.DeletionTimestamp.IsZero() {
+		found, err := r.hasOCSVolumes()
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		if found {
+			r.Log.Info("Found consumer PVs using default storageclasses, cannot proceed with uninstallation")
+			return ctrl.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
+		}
```
bindrad, line 134:
```suggestion
		if found, err := r.hasOCSVolumes(); err != nil {
			return ctrl.Result{}, err
		} else if found {
			r.Log.Info("Found consumer PVs using default storageclasses, cannot proceed with uninstallation")
			return ctrl.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
		}
```
Better to not pollute scope of variables

bindrad (changes requested):

controllers/datafoundationclient.go

controllers/datafoundationclient.go

bindrad (dismissed):

controllers/datafoundationclient.go
```diff
@@ -208,3 +234,44 @@ func (r *dataFoundationClientReconciler) reconcileDefaultFSStorageClassClaim() e
 
 	return nil
 }
+
+func (r *dataFoundationClientReconciler) hasOCSVolumes() (bool, error) {
+	storageClassList := &storagev1.StorageClassList{}
+	if err := r.list(storageClassList); err != nil {
+		return false, fmt.Errorf("failed to list Storage Classes: %v", err)
+	}
+	ocsStorageClass := make(map[string]bool)
```
nb-ohad, line 243: No need to use make here
```suggestion
	ocsStorageClass := map[string]bool{}
```

controllers/datafoundationclient.go
```diff
@@ -208,3 +234,44 @@ func (r *dataFoundationClientReconciler) reconcileDefaultFSStorageClassClaim() e
...
+	for i := range storageClassList.Items {
+		storageClassName := storageClassList.Items[i].Name
+		if storageClassName == defaultBlockStorageClassClaimName || storageClassName == defaultFSStorageClassClaimName {
+			ocsStorageClass[storageClassName] = true
+		}
+	}
```
nb-ohad, line 249: This entire loop does not make sense. In the end, you will get a map that only contains the following pairs:
```
defaultBlockStorageClassClaimName -> true
defaultFSStorageClassClaimName -> true
```

It is basically equivalent to 
```
ocsStorageClass[defaultBlockStorageClassClaimName] = true
ocsStorageClass[defaultFSStorageClassClaimName] = true
```

But there is an even larger issue here, it does NOT take into account any additional `storageclass`/`storageclassclaims` that the customer created (which are not the default ones)

The condition needs to be based on the Provisioner field of the storageclass not on the name.

controllers/datafoundationclient.go
```diff
@@ -115,6 +124,23 @@ func (r *dataFoundationClientReconciler) parseSpec(offering *v1alpha1.ManagedFus
 
 func (r *dataFoundationClientReconciler) reconcilePhases() (ctrl.Result, error) {
 	if !r.offering.DeletionTimestamp.IsZero() {
+		if found, err := r.hasOCSVolumes(); err != nil {
+			return ctrl.Result{}, err
+		} else if found {
+			r.Log.Info("Found consumer PVs using default storageclasses, cannot proceed with uninstallation")
```
nb-ohad, line 130: Why default? what if the customer has created an additional storageclass via a claim? And the message should be mentioning that we are addressing storageclasses that are provisioned by ODF. And don't use the "consumer" terminology anymore. 
```suggestion
			r.Log.Info("PVs provisioned by ODF  storage classes exists, cannot proceed with uninstallation")
```

nb-ohad (changes requested):

controllers/datafoundationclient.go

controllers/datafoundationclient.go
ezio-auditore: Fixed by checking provisioner name instead

controllers/datafoundationclient.go

controllers/datafoundationclient.go
```diff
@@ -248,3 +275,44 @@ func (r *dataFoundationClientReconciler) reconcileCSIKMSConnectionDetailsConfigM
...
+	}
+	ocsStorageClass := map[string]bool{}
+	for i := range storageClassList.Items {
+		storageClass := storageClassList.Items[i]
+		if storageClass.Provisioner == r.offering.Namespace+ocsCephFsProvisionerSuffix || storageClass.Provisioner == r.offering.Namespace+ocsRBDProvisionerSuffix {
+			ocsStorageClass[storageClass.Name] = true
+		}
+	}
+	pvList := &corev1.PersistentVolumeList{}
+	if err := r.list(pvList); err != nil {
```
nb-ohad, line 292: I believe that this will only list PVs from the set of namespaces the operator group allows for it. 
We need to list all PVs! for that I believe you can use the Unrestricted Client

controllers/datafoundationclient.go
```diff
@@ -248,3 +275,44 @@ func (r *dataFoundationClientReconciler) reconcileCSIKMSConnectionDetailsConfigM
...
+	if err := r.list(storageClassList); err != nil {
+		return false, fmt.Errorf("failed to list Storage Classes: %v", err)
+	}
+	ocsStorageClass := map[string]bool{}
+	for i := range storageClassList.Items {
+		storageClass := storageClassList.Items[i]
+		if storageClass.Provisioner == r.offering.Namespace+ocsCephFsProvisionerSuffix || storageClass.Provisioner == r.offering.Namespace+ocsRBDProvisionerSuffix {
+			ocsStorageClass[storageClass.Name] = true
+		}
+	}
```
nb-ohad, line 290: Why do we need to calculate the provisioners name on each loop iteration?
```suggestion
	ocsStorageClass := map[string]bool{}
	rbdProvisionerName = r.offering.Namespace + ocsRBDProvisionerSuffix
	fsProvisionerName := r.offering.Namespace + ocsCephFsProvisionerSuffix
	
	for i := range storageClassList.Items {
		storageClass := storageClassList.Items[i]
		if storageClass.Provisioner == rbdProvisionerName || storageClass.Provisioner == fsProvisionerName {
			ocsStorageClass[storageClass.Name] = true
		}
	}
```

nb-ohad (changes requested):

controllers/datafoundationclient.go

controllers/datafoundationclient.go
ezio-auditore: Updated to fetch using UnrestrictedClient

controllers/datafoundationclient.go
```diff
@@ -123,6 +133,23 @@ func (r *dataFoundationClientReconciler) parseSpec(offering *v1alpha1.ManagedFus
...
+			r.Log.Info("PVs provisioned by ODF  storage classes exists, cannot proceed with uninstallation")
+			return ctrl.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
+		}
+		r.Log.Info("issuing a delete for Default StorageClassClaims")
+		if err := r.delete(&r.defaultBlockStorageClassClaim); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to issue a delete for defaultBlockStorageClassClaim: %v", err)
+		}
+		if err := r.delete(&r.defaultFSStorageClassClaim); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to issue a delete for defaultFSStorageClassClaim: %v", err)
+		}
```
bindrad, line 148: Shouldn't we delete non-default storageClassClaims as well that are provisioned by `cephfs.csi.ceph.com` or `rbd.csi.ceph.com`?

controllers/datafoundationclient.go
```diff
@@ -248,3 +275,46 @@ func (r *dataFoundationClientReconciler) reconcileCSIKMSConnectionDetailsConfigM
...
+	if err := r.UnrestrictedClient.List(r.ctx, pvList); err != nil {
+		return false, fmt.Errorf("failed to list persistent volumes: %v", err)
+	}
+	for i := range pvList.Items {
+		scName := pvList.Items[i].Spec.StorageClassName
+		if ocsStorageClass[scName] {
+			return true, nil
+		}
+	}
+	return false, nil
```
bindrad, line 303: Shouldn't we list all the PVs and related storageClass that are preventing the uninstallation of the offering?
Because now customers can create multiple storageClassClaims so it will be difficult to track which PV is blocking the uninstallation of offering.

ezio-auditore: @bindrad So the ask is to list all the PVS blocking unistallation and log them?
Currently the code will map all the storageclasses provisioned using cephfs.csi.ceph.com or rbd.csi.ceph.com
later use this map to find if any pvs are using the said storageclasses.

bindrad: Yes, I think we should list all the PVs that are using ocsStorageClass as that is useful information to unblock uninstallation.

controllers/datafoundationclient.go
nb-ohad: That is a very good question! @ezio-auditore what do you think?

controllers/datafoundationclient.go
nb-ohad: Why should we log them? Who will read that log?

bindrad: Based on past experiences, what if SRE's don't have permission to list all the PVs in the cluster?
If we are logging the info, SRE can send the list to the customer asking to delete the PVs.

> put a potentially very big list of names in our log every few seconds

We will log the list only when uninstallation is triggered and if uninstallation is blocked because of a specific set of PVs in the cluster then I think that is a useful piece of information.

ezio-auditore: Sounds good. I will update the implementation to log the uninstallation blocking PVS

bindrad: After a discussion with Ohad regarding this, I think logging number of PVs should be good enough as there can be 100s of PVs so we will be logging a huge list. Instead SREs can view the PVs on cloud monitoring if required.

controllers/datafoundationclient.go
ezio-auditore: What I think
_If the user creates his custom storageClassClaim using the the default storageclient spec only then we need to delete those storageClassClaims,else we need to ignore it._

@nb-ohad @bindrad If this sounds ok I will implement the same here.

bindrad: > default storageclient spec

Are we even allowing customer to create non-default storageClient? Or do we have plans to allow customer to create non-default storageClients?

nb-ohad: > If the user creates his custom storageClassClaim using the the default storageclient spec
@ezio-auditore  that does not make any sense. The reason someone will create another claim is if they want some aspects of the storageclass to be different

ezio-auditore: @nb-ohad 
So why do we even need to remove the storageclassclaims created by users?

controllers/datafoundationclient.go
ezio-auditore: updatded with pvCount

controllers/datafoundationclient.go
nb-ohad: Because we are uninstalling the controller that reconciles it. That is precisely the gap that Dhruv has highlighted. 
But on the other hand, the CRD is not getting uninstalled and the CR themselves are not owned by our controller. 
So it is a gray area. But then again, we do remove the default storage class claims, so the behavior is not consistent.

I am not sure what are is best practice here

Madhu-1: Do you think it would be beneficial to prevent uninstallation until all user-created storage class claims have been deleted?

bindrad: According to me, the customers will be creating additional storageClassClaims to utilize storageClass with different settings. The expectation from the customer's side will be to have the storage running seamlessly. So we should block the uninstallation of the DFC offering if there are user-created storageClassClaims on the cluster.

@Madhu-1 @nb-ohad WDYT?

Madhu-1: Sounds good to me. This is the same as blocking the uninstall or reqeueing the request if there are still PV in the cluster.

nb-ohad: The existence of storageclassclaims or even storageclasses does not infer usage of storage. I am against blocking uninstallation based on that

controllers/datafoundationclient.go
```diff
@@ -248,3 +275,52 @@ func (r *dataFoundationClientReconciler) reconcileCSIKMSConnectionDetailsConfigM
...
+func (r *dataFoundationClientReconciler) hasOCSVolumes() (bool, error) {
+	storageClassList := &storagev1.StorageClassList{}
+	if err := r.list(storageClassList); err != nil {
+		return false, fmt.Errorf("failed to list Storage Classes: %v", err)
+	}
+	ocsStorageClass := map[string]bool{}
+	rbdProvisionerName := r.offering.Namespace + ocsRBDProvisionerSuffix
+	fsProvisionerName := r.offering.Namespace + ocsCephFsProvisionerSuffix
+	for i := range storageClassList.Items {
+		storageClass := storageClassList.Items[i]
```
nb-ohad, line 288: To prevent a copy operation, please take a reference
```suggestion
		storageClass := &storageClassList.Items[i]
```

controllers/datafoundationclient.go
```diff
@@ -248,3 +275,52 @@ func (r *dataFoundationClientReconciler) reconcileCSIKMSConnectionDetailsConfigM
...
 }
+
+func (r *dataFoundationClientReconciler) hasOCSVolumes() (bool, error) {
+	storageClassList := &storagev1.StorageClassList{}
+	if err := r.list(storageClassList); err != nil {
+		return false, fmt.Errorf("failed to list Storage Classes: %v", err)
+	}
+	ocsStorageClass := map[string]bool{}
+	rbdProvisionerName := r.offering.Namespace + ocsRBDProvisionerSuffix
+	fsProvisionerName := r.offering.Namespace + ocsCephFsProvisionerSuffix
```
nb-ohad, line 286: The `ocsRBDProvisionerSuffix` and `ocsCephFsProvisionerSuffix` are only used once here. Can we just embed the values?

controllers/datafoundationclient.go
```diff
@@ -248,3 +275,52 @@ func (r *dataFoundationClientReconciler) reconcileCSIKMSConnectionDetailsConfigM
...
+	for i := range storageClassList.Items {
+		storageClass := storageClassList.Items[i]
+		if storageClass.Provisioner == fsProvisionerName || storageClass.Provisioner == rbdProvisionerName {
+			ocsStorageClass[storageClass.Name] = true
+		}
+	}
+	pvList := &corev1.PersistentVolumeList{}
+	pvCount := 0
+	if err := r.UnrestrictedClient.List(r.ctx, pvList); err != nil {
```
nb-ohad, line 295: I believe we had an `r.UnrestrictedList(pvList)` operation.

controllers/datafoundationclient.go
```diff
@@ -248,3 +275,52 @@ func (r *dataFoundationClientReconciler) reconcileCSIKMSConnectionDetailsConfigM
 
 	return nil
 }
+
+func (r *dataFoundationClientReconciler) hasOCSVolumes() (bool, error) {
+	storageClassList := &storagev1.StorageClassList{}
+	if err := r.list(storageClassList); err != nil {
+		return false, fmt.Errorf("failed to list Storage Classes: %v", err)
+	}
+	ocsStorageClass := map[string]bool{}
```
nb-ohad, line 284:
```suggestion
	ocsStorageClasses := map[string]bool{}
```

controllers/datafoundationclient.go
```diff
@@ -248,3 +275,52 @@ func (r *dataFoundationClientReconciler) reconcileCSIKMSConnectionDetailsConfigM
...
+		if ocsStorageClass[scName] {
+			pvCount++
+		}
+	}
+	if pvCount > 0 {
+		r.Log.Info(fmt.Sprintf("Current count of PVs blocking uninstallation: %v", pvCount))
+		return true, nil
+	}
+
+	return false, nil
```
nb-ohad, line 309: It does not make sense to have logic/logging of messages base on the result of this func inside the func itself.
Let's have this function be `countOCSVolumes` which returns a number and handles the decision-making process outside.

main.go
```diff
@@ -133,9 +133,10 @@ func main() {
 	}
 
 	if err = (&controllers.ManagedFusionOfferingReconciler{
-		Client: mgr.GetClient(),
-		Log:    ctrl.Log.WithName("controllers").WithName("ManagedFusionOffering"),
-		Scheme: mgr.GetScheme(),
+		Client:             mgr.GetClient(),
+		UnrestrictedClient: getUnrestrictedClient(),
```
nb-ohad, line 137: Can we just create the unrestricted client once and rescue it in all controllers?

nb-ohad (changes requested):

controllers/datafoundationclient.go
bindrad: @Madhu-1 Will the storageClassClaims get removed if the storageClient CR/client-operator is removed?

controllers/datafoundationclient.go
```diff
@@ -34,13 +37,20 @@ type dataFoundationClientReconciler struct {
...
+	defaultStorageClientName             = "ocs-storageclient"
+	ocsCephFsProvisionerSuffix           = ".cephfs.csi.ceph.com"
+	ocsRBDProvisionerSuffix              = ".rbd.csi.ceph.com"
 )
 
 //+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclients,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclassclaims,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=operators.coreos.com,resources=clusterserviceversions,verbs=get;list;watch;delete;update;patch
+//+kubebuilder:rbac:groups="",resources=persistentvolumes,verbs=get;list;watch
+//+kubebuilder:rbac:groups="storage.k8s.io",resources=storageclass,verbs=get;list;watch
```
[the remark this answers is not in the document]
bindrad: We need access to get, list and watch storageClass to fetch the PVs that are using ocs storageclasses

controllers/datafoundationclient.go
```diff
@@ -34,13 +37,20 @@ type dataFoundationClientReconciler struct {
...
+	defaultFSStorageClassClaimName       = "ocs-storagecluster-cephfs"
+	defaultStorageClientName             = "ocs-storageclient"
+	ocsCephFsProvisionerSuffix           = ".cephfs.csi.ceph.com"
+	ocsRBDProvisionerSuffix              = ".rbd.csi.ceph.com"
 )
 
 //+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclients,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclassclaims,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=operators.coreos.com,resources=clusterserviceversions,verbs=get;list;watch;delete;update;patch
+//+kubebuilder:rbac:groups="",resources=persistentvolumes,verbs=get;list;watch
```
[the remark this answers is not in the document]
bindrad: Yes, it gives us enough privileges

controllers/datafoundationclient.go
```diff
@@ -75,12 +85,12 @@ func (r *dataFoundationClientReconciler) initReconciler(reconciler *ManagedFusio
...
 
-	r.storageClient.Name = "ocs-storageclient"
+	r.storageClient.Name = defaultStorageClientName
 	r.storageClient.Namespace = offering.Namespace
 
-	r.defaultBlockStorageClassClaim.Name = "ocs-storagecluster-ceph-rbd"
+	r.defaultBlockStorageClassClaim.Name = defaultBlockStorageClassClaimName
 
-	r.defaultFSStorageClassClaim.Name = "ocs-storagecluster-cephfs"
+	r.defaultFSStorageClassClaim.Name = defaultFSStorageClassClaimName
```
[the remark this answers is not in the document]
bindrad: These are the default storageClassClaims that the agent is creating.

controllers/datafoundationclient.go
Madhu-1: If there are any PVCs utilizing the storageclass of SCC, it cannot be removed. Similarly, the StorageClient cannot be deleted until all the SCCs using it are deleted.

Madhu-1: I forgot to post my comment on this one :facepalm: 

Instead of checking PV, it would be better to check SCC before deleting the storageClient for two reasons. Firstly, we are creating StorageClient and StorageClassClaims, so it makes sense to check for the resources we have created. Secondly, listing PVs can be a heavy operation compared to listing SCCs, as there are very few.

nb-ohad: @Madhu-1 We must check for PVs, that is not even a question. We need to be sure that no-one is consuming any storage from our provider before we delete the client. We should not care how the storageclass, backing the PV, was created (manually or via  SCC). 

The question here was about SCC, I think we should just delete the lot of them

Madhu-1: @nb-ohad, the ocs-client-operator is already checking for any PV before deleting the SCC. Why do we need to recheck the PV here again?

>The question here was about SCC, I think we should just delete the lot of them

You mean list all the SCC, delete them, and wait for them to be deleted before deleting the storage client right?

Madhu-1: Do we support Users creating custom storageclass and static provisioning without SCC in managed service?

bindrad: > Do we support Users creating custom storageclass and static provisioning without SCC in managed service?

On consumer clusters, customer can do anything they want to.

bindrad: > You mean list all the SCC, delete them, and wait for them to be deleted before deleting the storage client right?

I think we can issue a delete for all SCCs and then issue a delete for storageClient. No need to wait for the removal of all SCCs before deleting storageClient because client-operator won't delete storageClient unless all SCCs are removed.

nb-ohad: @Madhu-1 I am aware the ocs-client-operator is already checking for that. 
But a k8s delete operation is unrevertable so we check for potential lock conditions, from the operator that owns the lifecycle of the offering (in this case the client operator) before issuing the delete request.

rchikatw:
As per discussion with Dhruv closing this PR as the same work is being addressed in a different [PR](https://github.com/red-hat-storage/managed-fusion-agent/pull/63) and Dhruv is working on it.

ezio-auditore:
closing since another PR #63 already addressing this

PR #58, opened 2023-05-11
Update ManagedFusionOffering API to use yaml instead of map for offering configuration

ezio-auditore:
Currently ManagedFusionOffering API uses a map for the offering variables, this pr enables to use yaml instead of map in the ManagedFusionOffering CR


controllers/datafoundation.go
```diff
@@ -202,11 +203,15 @@ func (r *dataFoundationReconciler) reconcilePhases() (ctrl.Result, error) {
 
 func (r *dataFoundationReconciler) parseSpec(offering *v1alpha1.ManagedFusionOffering) error {
 	r.Log.Info("Parsing ManagedFusionOffering Data Foundation spec")
-
+	config := make(map[string]string)
+	err := yaml.Unmarshal([]byte(offering.Spec.Config), &config)
+	if err != nil {
+		r.Log.Error(err, "an error occurred while unmarshaling the YAML config")
+		return fmt.Errorf("failed to unmarshal YAML config")
+	}
```
bindrad, line 211: We should unmarshal into a struct as we already know what should be passed in the config section of MFO CR

controllers/datafoundationclient.go
```diff
@@ -80,9 +81,15 @@ func (r *dataFoundationClientReconciler) initReconciler(reconciler *ManagedFusio
 
 func (r *dataFoundationClientReconciler) parseSpec(offering *v1alpha1.ManagedFusionOffering) error {
 	r.Log.Info("Parsing ManagedFusionOffering Data Foundation Client spec")
+	config := make(map[string]string)
+	err := yaml.Unmarshal([]byte(offering.Spec.Config), &config)
+	if err != nil {
+		r.Log.Error(err, "an error occurred while unmarshaling the YAML config")
+		return fmt.Errorf("failed to unmarshal YAML config")
+	}
```
bindrad, line 89: We should unmarshal into a struct as we already know what should be passed in the config section of MFO CR

bindrad (changes requested):
You need to check if the config section has a valid yaml or not in the managedFusionOffering reconciler(before going into reconciliation of offerings).

bindrad:
@ezio-auditore Can you please update the PR description as well?

controllers/datafoundation.go
ezio-auditore: updated to using struct

controllers/datafoundationclient.go
ezio-auditore: updated to using struct

controllers/datafoundation.go
```diff
@@ -203,26 +203,22 @@ func (r *dataFoundationReconciler) reconcilePhases() (ctrl.Result, error) {
...
 	isValid := true
-	var err error
-	var usableCapacityInTiB int
-	usableCapacityInTiBAsString, found := offering.Spec.Config["usableCapacityInTiB"]
-	if !found {
+	if dfSpec.usableCapacityInTiB == 0 {
```
ezio-auditore, line 213: I am a bit sceptical about this. any suggestions?

bindrad: 0 is the default value, right? 
If nothing is passed then 0 will be set, right?

controllers/datafoundation.go
```diff
@@ -203,26 +203,22 @@ func (r *dataFoundationReconciler) reconcilePhases() (ctrl.Result, error) {
 func (r *dataFoundationReconciler) parseSpec(offering *v1alpha1.ManagedFusionOffering) error {
 	r.Log.Info("Parsing ManagedFusionOffering Data Foundation spec")
 
+	var dfSpec dataFoundationSpec
+	err := yaml.Unmarshal([]byte(offering.Spec.Config), &dfSpec)
+	if err != nil {
+		r.Log.Error(err, "an error occurred while unmarshaling the YAML config")
+		return fmt.Errorf("failed to unmarshal YAML config")
+	}
```
bindrad, line 211:
```suggestion
	if err := yaml.Unmarshal([]byte(offering.Spec.Config), &dfSpec); err != nil {
		r.Log.Error(err, "an error occurred while unmarshaling the YAML config")
		return fmt.Errorf("failed to unmarshal YAML config")
	}
```

controllers/datafoundationclient.go
```diff
@@ -89,18 +90,23 @@ func (r *dataFoundationClientReconciler) initReconciler(reconciler *ManagedFusio
 func (r *dataFoundationClientReconciler) parseSpec(offering *v1alpha1.ManagedFusionOffering) error {
 	r.Log.Info("Parsing ManagedFusionOffering Data Foundation Client spec")
 
+	var dfcSpec dataFoundationClientSpec
+	err := yaml.Unmarshal([]byte(offering.Spec.Config), &dfcSpec)
+	if err != nil {
+		r.Log.Error(err, "an error occurred while unmarshaling the YAML config")
+		return fmt.Errorf("failed to unmarshal YAML config")
+	}
```
bindrad, line 98:
```suggestion
	if err := yaml.Unmarshal([]byte(offering.Spec.Config), &dfcSpec); err != nil {
		r.Log.Error(err, "an error occurred while unmarshaling the YAML config")
		return fmt.Errorf("failed to unmarshal YAML config")
	}
```

bindrad:
Please update the description

ezio-auditore:
> You need to check if the config section has a valid yaml or not in the managedFusionOffering reconciler(before going into reconciliation of offerings).

But the config section will have two different structs based on the offering. Do you want me to validate based off the offering in managedFusionOffering reconciler?

controllers/datafoundation.go
ezio-auditore: updated

controllers/datafoundationclient.go

bindrad:
> > You need to check if the config section has a valid yaml or not in the managedFusionOffering reconciler(before going into reconciliation of offerings).
> 
> But the config section will have two different structs based on the offering. Do you want me to validate based off the offering in managedFusionOffering reconciler?

In the managedFusionOffering reconciler, you just need to validate if the config is a valid yaml or not. To do this, you don't need to know what's the structure of yaml.

ezio-auditore:
> > > You need to check if the config section has a valid yaml or not in the managedFusionOffering reconciler(before going into reconciliation of offerings).
> > 
> > 
> > But the config section will have two different structs based on the offering. Do you want me to validate based off the offering in managedFusionOffering reconciler?
> 
> In the managedFusionOffering reconciler, you just need to validate if the config is a valid yaml or not. To do this, you don't need to know what's the structure of yaml.

Updated in managedFusionOffering reconciler

controllers/datafoundation.go
ezio-auditore: @bindrad if the field is present in the yaml string but it's value is empty or an invalid integer it throws an error
if field itself is not present it defaults to 0

controllers/datafoundation.go
```diff
@@ -203,26 +203,21 @@ func (r *dataFoundationReconciler) reconcilePhases() (ctrl.Result, error) {
 func (r *dataFoundationReconciler) parseSpec(offering *v1alpha1.ManagedFusionOffering) error {
 	r.Log.Info("Parsing ManagedFusionOffering Data Foundation spec")
 
+	var dfSpec dataFoundationSpec
+	if err := yaml.Unmarshal([]byte(offering.Spec.Config), &dfSpec); err != nil {
+		r.Log.Error(err, "an error occurred while unmarshaling the YAML config")
+		return fmt.Errorf("failed to unmarshal YAML config")
```
bindrad, line 209:
```suggestion
		return fmt.Errorf("failed to unmarshal YAML offering config: %v", err)
```

controllers/datafoundationclient.go
```diff
@@ -89,18 +90,22 @@ func (r *dataFoundationClientReconciler) initReconciler(reconciler *ManagedFusio
 func (r *dataFoundationClientReconciler) parseSpec(offering *v1alpha1.ManagedFusionOffering) error {
 	r.Log.Info("Parsing ManagedFusionOffering Data Foundation Client spec")
 
+	var dfcSpec dataFoundationClientSpec
+	if err := yaml.Unmarshal([]byte(offering.Spec.Config), &dfcSpec); err != nil {
+		r.Log.Error(err, "an error occurred while unmarshaling the YAML config")
+		return fmt.Errorf("failed to unmarshal YAML config")
```
bindrad, line 96:
```suggestion
		return fmt.Errorf("failed to unmarshal YAML offering config: %v", err)
```

controllers/managedfusionoffering_controller.go
```diff
@@ -180,6 +181,9 @@ func (r *ManagedFusionOfferingReconciler) reconcilePhases() (reconcile.Result, e
 		if err := r.reconcileSubscription(); err != nil {
 			return ctrl.Result{}, err
 		}
+		if err := reconcileOfferingConfig(r.managedFusionOffering); err != nil {
+			return ctrl.Result{}, err
+		}
```
bindrad, line 186: The function shouldn't be named `reconcile*` as we are not reconciling anything in the function. Why are we calling it after creating operatorGroup, CatSrc and Subscription?
The config should be validated just after we get ManagedFusionOffering CR. According to me, if the config is invalid then there is no point in creating operatorGroup, catSrc and Subscription.

controllers/managedfusionoffering_controller.go
```diff
@@ -190,6 +194,15 @@ func (r *ManagedFusionOfferingReconciler) reconcilePhases() (reconcile.Result, e
 
 }
 
+func reconcileOfferingConfig(offering *v1alpha1.ManagedFusionOffering) error {
+
+	offeringConfigYaml := map[interface{}]interface{}{}
```
bindrad, line 199: Just for my understanding, why `map[interface{}]interface{}{}` and not `interface{}`?

bindrad (changes requested):

controllers/datafoundation.go
ezio-auditore: updated

controllers/datafoundationclient.go

controllers/managedfusionoffering_controller.go
ezio-auditore: renamed func to validateOfferingConfig and set the calling before operatorGroup

controllers/managedfusionoffering_controller.go
ezio-auditore: My bad, I din't realise we are not going to parse the spec here or iterate within the elements  but only validating the yaml.
Changed the implementation to use interface{}

controllers/managedfusionoffering_controller.go
```diff
@@ -190,6 +194,15 @@ func (r *ManagedFusionOfferingReconciler) reconcilePhases() (reconcile.Result, e
...
 }
 
+func validateOfferingConfig(offering *v1alpha1.ManagedFusionOffering) error {
+
+	var offeringConfigYaml interface{}
+	if err := yaml.Unmarshal([]byte(offering.Spec.Config), &offeringConfigYaml); err != nil {
+		return fmt.Errorf("invalid yaml found in ManagedFusionOffering config: %v", err)
+	}
+	return nil
+}
```
nb-ohad, line 204: Why do we need to validate here and then unmarshal again in the offering code? 
Can we just unmarshal once just before pluginReconcile? 

The way I would do that is by asking the offering code to provide an empty instance of the offeringConfig struct as an interface. Unmarshl into that interface, using `StrictUnmarshal` (this will cover both YAML validity check and specific offering validity check) then pass that into to the offering reconcile code. 

The offering can then use `config.(OfferingConfigStructType)` to case it into the correct type and use that

nb-ohad (changes requested):

controllers/managedfusionoffering_controller.go
rewantsoni: @nb-ohad I tried unmarshalling it into a interface and type cast it using config.(OfferingConfigStructType), it's throwing a error.

We can unmarshall into a interface or map[sting]interface{} and pass that into offering reconcile code and then parseSpec can use that interface to read values

bindrad:
Closing this PR as the work was done in https://github.com/red-hat-storage/managed-fusion-agent/pull/62

PR #61, opened 2023-05-18
Add namespace selector in prometheus CR for certain resources and add missing labels on service monitor

bindrad:
namespace selector is added for
- prometheusRules
- serviceMonitor
- podMonitor

If the namespace selector is not passed then the own namespace is considered so now, Prometheus will select the resources having the label `"app": "managed-fusion-agent"` from all the namespaces.

Mon label was missing from this service monitor so Prometheus was not able to discover this service monitor which resulted in missing metrics.

### DIFF
--- a/controllers/managedfusion_controller.go
+++ b/controllers/managedfusion_controller.go
@@ -709,6 +709,7 @@ func (r *ManagedFusionReconciler) reconcileK8SMetricsServiceMonitor() error {
 		}
 		desired := templates.K8sMetricsServiceMonitorTemplate.DeepCopy()
 		r.k8sMetricsServiceMonitor.Spec = desired.Spec
+		utils.AddLabel(r.k8sMetricsServiceMonitor, monLabelKey, monLabelValue)
 		return nil
 	})
 	if err != nil {
--- a/templates/prometheus.go
+++ b/templates/prometheus.go
@@ -35,6 +35,13 @@ var resourceSelector = metav1.LabelSelector{
 	},
 }
 
+var namespaceSelector = metav1.LabelSelector{
+	MatchExpressions: []metav1.LabelSelectorRequirement{{
+		Key:      "kubernetes.io/metadata.name",
+		Operator: metav1.LabelSelectorOpExists,
+	}},
+}
+
 const (
 	//TODO: Construct monitoring dns name dynamically based on AWS and IBM Cloud Monitoring regions
 	RemoteWriteDNSName                  string = "ingest.us-south.monitoring.cloud.ibm.com"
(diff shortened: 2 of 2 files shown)


controllers/managedfusion_controller.go
```diff
@@ -709,6 +709,7 @@ func (r *ManagedFusionReconciler) reconcileK8SMetricsServiceMonitor() error {
 		}
 		desired := templates.K8sMetricsServiceMonitorTemplate.DeepCopy()
 		r.k8sMetricsServiceMonitor.Spec = desired.Spec
+		utils.AddLabel(r.k8sMetricsServiceMonitor, monLabelKey, monLabelValue)
```
leelavg, line 712: - why was this not required in deployer but needed in agent?

bindrad: We are doing the same in deployer but differently. 
Checkout reconcileMonitoringResources function, we do add labels to podMonitors, serviceMonitors and prometheusRules. In Agent, we moved that piece of code in dataFoundation reconciler and didn't think that this is also required for the k8s-metric-service-monitor.
Ref: https://github.com/red-hat-storage/ocs-osd-deployer/blob/main/controllers/managedocs_controller.go#L1395-L1435

leelavg: I meant the label (app: ...-agent) for service monitor as I was under the impression service monitor doesn't require to be labelled.

let me know if I'm wrong.

nb-ohad: The prom instance only selects resources that have a label matching the label selector. 
This is how we configure our prom the select monitoring resources.

bindrad: @leelavg we have labelSelector for serviceMonitor, podMonitor and promRules in prometheus CR. So our prometheus instance selects the resources based on label.

templates/prometheus.go
```diff
@@ -35,6 +35,12 @@ var resourceSelector = metav1.LabelSelector{
 	},
 }
 
+var namespaceSelector = metav1.LabelSelector{
+	MatchLabels: map[string]string{
+		"misf.ibm.com/managed": "true",
```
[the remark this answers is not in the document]
bindrad: > Who is adding this label to every offering namespace?

When the namespace is created by lambda we add `"misf.ibm.com/managed": "true"` label.

> can we configure prom to look at all monitoring resources labeled with our label inside all namespaces?

Yes, I think we should be able to do that.

nb-ohad:
@bindrad can you please update the subject line to reflect the changes

templates/prometheus.go
```diff
@@ -35,6 +35,14 @@ var resourceSelector = metav1.LabelSelector{
...
 }
 
+var namespaceSelector = metav1.LabelSelector{
+	MatchExpressions: []metav1.LabelSelectorRequirement{{
+		Key:      "kubernetes.io/metadata.name",
+		Operator: metav1.LabelSelectorOpExists,
+		Values:   []string{},
+	}},
+}
+
```
nb-ohad, line 45: Why not?
```suggestion
var namespaceSelector = metav1.LabelSelector{}
```

bindrad: I tried doing that but the `ServiceMonitorNamespaceSelector`, `PodMonitorNamespaceSelector` and `RuleNamespaceSelector` treat is as an empty value. So only the own namespace is selected.

leelavg: - `..opExist` is also a valid identifier with no value but a valid key
- not sure why we want to have empty struct (just curious about comment reasoning)

bindrad: > ..opExists is also a valid identifier with no value but a valid key

In the go API doc, this is mentioned `If the operator is Exists or DoesNotExist, the values array must be empty`. So I used an empty array instead of no value. I will try again with passing no value.

> not sure why we want to have empty struct (just curious about comment reasoning)

I guess Ohad mentioned it because of this `An empty label selector matches all objects. A null label selector matches no objects.`. This statement is mentioned in the go API doc of the label selector.
https://pkg.go.dev/k8s.io/apimachinery@v0.26.1/pkg/apis/meta/v1#LabelSelector

leelavg: > I will try again with passing no value.
- ack, iirc, I used that during pricing
- thanks for the other ref

bindrad: @leelavg Worked without passing value, thanks!

nb-ohad (approved):

PR #63, opened 2023-05-23
Uninstallation flow for DFC offering

bindrad:
Raised this PR as replacement of https://github.com/red-hat-storage/managed-fusion-agent/pull/55

This PR implements uninstallation flow for DFC offering
- checks for PVs using ocs storage classes and halt uninstallation if found
- issue delete for all storageclass claims
- issue delete for storageclient

### DIFF
--- a/controllers/datafoundationclient.go
+++ b/controllers/datafoundationclient.go
@@ -3,13 +3,16 @@ package controllers
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	v1alpha1 "github.com/red-hat-storage/managed-fusion-agent/api/v1alpha1"
 	"github.com/red-hat-storage/managed-fusion-agent/utils"
 	ocsclient "github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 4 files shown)


controllers/datafoundationclient.go
```diff
@@ -40,11 +43,16 @@ const (
 	csiKMSConnectionDetailsConfigMapName = "csi-kms-connection-details"
 	storageClientCRDName                 = "storageclients.ocs.openshift.io"
 	storageClassClaimCRDName             = "storageclassclaims.ocs.openshift.io"
+	defaultStorageClientName             = "ocs-storageclient"
```
nb-ohad, line 46: Why default? this is the only storageclient we support right now. 
```suggestion
	storageClientName             = "ocs-storageclient"
```

controllers/datafoundationclient.go
```diff
@@ -124,6 +132,29 @@ func (r *dataFoundationClientReconciler) reconcilePhases() (ctrl.Result, error)
...
+		if err := r.unrestrictedList(storageClassClaimList); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to list storage class claims: %v", err)
+		}
+		for i := range storageClassClaimList.Items {
+			storageClassClaim := &storageClassClaimList.Items[i]
+			r.Log.Info("Issuing a delete for storageClassClaim %s", storageClassClaim.Name)
+			if err := r.unrestrictedDelete(storageClassClaim); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to issue a delete for storageClassClaim %s: %v", storageClassClaim.Name, err)
+			}
+		}
```
nb-ohad, line 152: Two issues I see with this loop.
1. if one SCC delete fails, you will not issue and delete for anything else after it.
2. You might be issuing a delete for an already "marked for deletion" SCC again and again and again.

controllers/datafoundationclient.go
```diff
@@ -124,6 +132,29 @@ func (r *dataFoundationClientReconciler) reconcilePhases() (ctrl.Result, error)
...
+
+		r.Log.Info("issuing a delete for StorageClient")
+		if err := r.delete(&r.storageClient); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to issue a delete for Storage Client: %v", err)
+		}
```
nb-ohad, line 157: Do we need this? we are the controller owner of this resource. It will get deleted when the offering CR will get deleted. Am I wrong here?

nb-ohad (changes requested):

controllers/datafoundationclient.go
bindrad: I will try to explain using a scenario
- Let's assume that the ocs-client operator adds a finalizer to a secret for some reason if a storageClient is created.
- This finalizer should be removed by the ocs-client operator if storageClient is in deleting state.

Now, if we delete offering CR that is the owner of storageClient then storageClient will be removed instantly as the owner(offering CR) was removed. At this moment, it is possible that ocs-client operator didn't trigger a reconcile quick enough to remove the finalizer from the secret. This will block deletion of namespace. So according to me, the agent should initiate the deletion of storageClient and after the successful removal of storageClient, we should remove offering CR.

bindrad: We do the same for storageCluster in DF offering

nb-ohad: What you are describing is a bug in the uninstallation/deletion flow of the client. If the client added a finalizer on a resource it needs to act in one of two consistent ways:
1. remove the finalizer if the client is marked for delete
2. stop uninstallation until the secret is removed. 
Which one is dependent on what it tries to achieve.

Any other behavior will be considered a bug and the client developers will need to fix it.

nb-ohad: I believe we do the same because we copy-pasted the current behavior form the old deployer where there was no governing offering CR that held an owner ref. We need to evaluate if a change is also needed there.

bindrad: > where there was no governing offering CR that held an owner ref

We had managedOCS CR owning the storageCluster, right?

> eventually consistent model for offering removal or a sync one

Can you elaborate on this?

controllers/datafoundationclient.go
bindrad: updated

controllers/datafoundationclient.go
bindrad: updated

controllers/datafoundationclient.go
bindrad: For now, we can follow the behaviour of the deployer. If required, we can evaluate if a change is required for both offerings. I have updated the PR to follow an eventually consistent model for offering removal.

PR #64, opened 2023-05-29
Fix MFO update API

rewantsoni:
1. Export the variables so that they can be unmarshalled
2. Pass the pointer while typecasting

### DIFF
--- a/config/samples/misf_v1alpha1_managedfusionoffering.yaml
+++ b/config/samples/misf_v1alpha1_managedfusionoffering.yaml
@@ -6,6 +6,6 @@ metadata:
 spec:
   kind: DF
   release: "4.12"
-  config: 
-    usableCapacityInTiB: "4"
+  config: |
+    usableCapacityInTiB: 4
     onboardingValidationKey: MIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIICCgKCAgEAshEZLax4CqF6J1rgvBcjK9wVpM9BGxeJXy5LCXL39jkTkn5Nu2Fdxjo90ppycYtrfMXXF3KZGcw75LfB+vLa7P7Q1bjT6B8mn4drZfDZJ4CGS++FUbA3NpyhdGrwB0rGmXOfF0NwhKZM+lkCAwEAAQ==
--- a/controllers/datafoundation.go
+++ b/controllers/datafoundation.go
@@ -50,15 +50,15 @@ const (
 )
 
 type dataFoundationSpec struct {
-	usableCapacityInTiB     int
-	onboardingValidationKey string
+	UsableCapacityInTiB     int
+	OnboardingValidationKey string
 }
 
 type dataFoundationReconciler struct {
 	*ManagedFusionOfferingReconciler
 
 	offering                       *v1alpha1.ManagedFusionOffering
-	spec                           dataFoundationSpec
+	spec                           *dataFoundationSpec
 	onboardingValidationKeySecret  corev1.Secret
 	storageCluster                 ocsv1.StorageCluster
 	cephIngressNetworkPolicy       netv1.NetworkPolicy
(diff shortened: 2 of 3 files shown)


controllers/datafoundation.go
```diff
@@ -128,7 +128,7 @@ func dfReconcile(offeringReconciler *ManagedFusionOfferingReconciler, offering *
 	// Set up the datafoundation sub-reconciler
 	r := dataFoundationReconciler{}
 	r.initReconciler(offeringReconciler, offering)
-	r.spec = offeringSpec.(dataFoundationSpec)
+	r.spec = *offeringSpec.(*dataFoundationSpec)
```
bindrad, line 131:
```suggestion
	r.spec, ok = *offeringSpec.(*dataFoundationSpec)
	if !ok {
	    return ctrl.Result{}, fmt.Errorf("Error while typecasting into dataFoundationSpec)
	 }	
```
I think we should catch the error and return otherwise the code will panic. Please feel free to update the error message to something more meaningful. Same comment for `dfcReconcile`

bindrad (changes requested):

controllers/datafoundation.go
rewantsoni: @nb-ohad suggested that if there is an error typecasting we should panic

bindrad: And what was the reason for that?
If we panic then the pod will go into CLBO, if we don't then we will have a proper error mentioning there is something wrong with the `spec.config` section.

rewantsoni: Yes, I agree having a error message instead of code panic and pod running to CLBO will be better

nb-ohad: @bindrad 
if there is a panic here it means that the developer made the mistake of using one type in one place of the code while using another type here. This is a coding mistake, not a runtime error. If we could catch that in compile or lint I would prefer that, but as we cannot, the next best thing is a panic as it is the most obvious way to get the developer's attention, and will be consistent on every invocation!

rewantsoni: updated

bindrad: Okay, got it. Thanks!

config/samples/misf_v1alpha1_managedfusionoffering.yaml
```diff
@@ -6,6 +6,6 @@ metadata:
 spec:
   kind: DF
   release: "4.12"
-  config: 
+  config: |
     usableCapacityInTiB: "4"
```
bindrad, line 10:
```suggestion
    usableCapacityInTiB: 4
```
usableCapacityInTiB's value should be an int and not a string

bindrad (changes requested):

bindrad (dismissed):

leelavg (dismissed):

controllers/datafoundation.go
nb-ohad, line 131: I believe this is wrong. you are creating a copy of the `dataFoundationSpec` and then a copy of the `offeringSpec`. I believe the issue here is that `r.spec` is not a pointer (ref type)

controllers/datafoundationclient.go
```diff
@@ -80,7 +80,7 @@ func dfcGetOfferingSpecInstance() *dataFoundationClientSpec {
 func dfcReconcile(offeringReconciler *ManagedFusionOfferingReconciler, offering *v1alpha1.ManagedFusionOffering, offeringSpec interface{}) (ctrl.Result, error) {
 	r := dataFoundationClientReconciler{}
 	r.initReconciler(offeringReconciler, offering)
-	r.spec = offeringSpec.(dataFoundationClientSpec)
+	r.spec = *offeringSpec.(*dataFoundationClientSpec)
```
nb-ohad, line 83: I believe this is wrong. you are creating a copy of the `dataFoundationClientSpec` and then a copy of the `offeringSpec`. I believe the issue here is that `r.spec` is not a pointer (ref type)

nb-ohad (changes requested):

bindrad (approved):

nb-ohad (approved):

PR #67, opened 2023-06-06
Update the mfo api to have all the fields as required

rewantsoni:
Remove omitempty from all the fields to make them required. When we apply the CR for the offering with one or more fields not there, it will fail due to missing fields.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2208167

### DIFF
--- a/api/v1alpha1/managedfusionoffering_types.go
+++ b/api/v1alpha1/managedfusionoffering_types.go
@@ -30,9 +30,9 @@ const (
 
 // ManagedFusionOfferingSpec defines the desired state of ManagedFusionOffering
 type ManagedFusionOfferingSpec struct {
-	Kind    OfferingKind `json:"kind,omitempty"`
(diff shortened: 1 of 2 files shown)


bindrad:
Please add one line description along with BZ link

rewantsoni:
> Please add one line description along with BZ link

@bindrad updated the PR

bindrad:
Should we explicitly mention that the fields are required using kubebuilder markers?
`// +kubebuilder:validation:Required`

rewantsoni:
> Should we explicitly mention that the fields are required using kubebuilder markers? `// +kubebuilder:validation:Required`

I don't think we require the markers explicitly, the CRD with and without the markers remains the same

bindrad (approved):

leelavg (approved):

nb-ohad (approved):

PR #70, opened 2023-06-29
Handle 1-4TiB provider config and upscaling

raaizik:
Resolves [Faas-44](https://jsw.ibm.com/browse/FAAS-44), [Faas-45](https://jsw.ibm.com/browse/FAAS-45):

- Adds support for `UsableCapacityInTiB` config knob with values 1-4TiB (units included)
- Performs up-scaling and prevents down-scaling

TODOs:
- [ ] Check regarding optimal perf of different setups of {capacity, OSD size} combos 
- [x] Check use case of non-integer `UsableCapacityInTiB` is handled correctly
- [x] Test manually

image

### DIFF
--- a/controllers/datafoundation.go
+++ b/controllers/datafoundation.go
@@ -283,7 +283,7 @@ func (r *dataFoundationReconciler) reconcileStorageCluster() error {
 		}
 
 		// Convert the desired size to the device set count based on the underlying OSD size
-		desiredDeviceSetCount := int(math.Ceil(UsableCapacityInTiB.AsApproximateFloat64() / desiredOSDSizeInTiB.AsApproximateFloat64()))
+		desiredDeviceSetCount := int(math.Ceil(float64(UsableCapacityInTiB.ScaledValue(resource.Tera)) / float64(desiredOSDSizeInTiB.ScaledValue(resource.Tera))))
 
 		// Get the desired storage device set from storage cluster template
 		sc := templates.StorageClusterTemplate.DeepCopy()
--- a/controllers/datafoundation.go
+++ b/controllers/datafoundation.go
@@ -51,7 +51,7 @@ const (
 )
 
 type dataFoundationSpec struct {
-	UsableCapacityInTiB     resource.Quantity
+	UsableCapacityInTiB     int
 	OnboardingValidationKey string
 }
 
@@ -263,10 +263,7 @@ func (r *dataFoundationReconciler) reconcileOnboardingValidationSecret() error {
 func (r *dataFoundationReconciler) reconcileStorageCluster() error {
 	r.Log.Info("Reconciling StorageCluster")
 
-	UsableCapacityInTiB := r.spec.UsableCapacityInTiB
-	OSDSizeUpperBoundInTiB := templates.OSDSizeUpperBoundInTib
-
-	if UsableCapacityInTiB.CmpInt64(0) <= 0 {
+	if r.spec.UsableCapacityInTiB <= 0 {
 		return fmt.Errorf("invalid or missing field: UsableCapacityInTiB, provided value is %v", r.spec.UsableCapacityInTiB)
 	}
 
@@ -275,11 +272,14 @@ func (r *dataFoundationReconciler) reconcileStorageCluster() error {
 			return err
 		}
 
+		UsableCapacityInTiB := resource.MustParse(fmt.Sprintf("%dTi", r.spec.UsableCapacityInTiB))
+		OSDSizeUpperBoundInTiB := resource.MustParse(fmt.Sprintf("%dTi", templates.OSDSizeUpperBoundInTiB))
+
 		// Get the desired capacity in TiB planned to be consumed by user in range 1-4TiB
 		// Set desired to be the minimum
 		desiredOSDSizeInTiB := UsableCapacityInTiB
 		if OSDSizeUpperBoundInTiB.Cmp(UsableCapacityInTiB) < 0 {
-			desiredOSDSizeInTiB = *OSDSizeUpperBoundInTiB
+			desiredOSDSizeInTiB = OSDSizeUpperBoundInTiB
 		}
 
 		// Convert the desired size to the device set count based on the underlying OSD size
--- a/controllers/datafoundation.go
+++ b/controllers/datafoundation.go
@@ -16,7 +16,9 @@ package controllers
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"math"
+	"strconv"
 	"strings"
 	"time"
 
@@ -271,12 +273,21 @@ func (r *dataFoundationReconciler) reconcileStorageCluster() error {
 			return err
 		}
 
-		// Convert the desired size to the device set count based on the underlaying OSD size
-		desiredDeviceSetCount := int(math.Ceil(float64(r.spec.UsableCapacityInTiB) / templates.OSDSizeInTiB))
+		// get the desired capacity in TiB planned to be consumed by user in range 1-4TiB
+		desiredOSDSizeInTiB := math.Min(math.Ceil(float64(r.spec.UsableCapacityInTiB)), templates.VerticalScalerUpperBoundInTiB)
 
-		// Get the storage device set count of the current storage cluster
+		// Convert the desired size to the device set count based on the underlying OSD size
+		desiredDeviceSetCount := int(math.Ceil(float64(r.spec.UsableCapacityInTiB)) / desiredOSDSizeInTiB)
+		// Get the storage device set count and OSD size of the current storage cluster
 		currDeviceSetCount := 0
+		currOSDSizeInTiB := 0.0
 		if desiredStorageDeviceSet := findStorageDeviceSet(r.storageCluster.Spec.StorageDeviceSets, defaultDeviceSetName); desiredStorageDeviceSet != nil {
+			// Avoid overflow
+			if len(strconv.Itoa(int(desiredStorageDeviceSet.DataPVCTemplate.Spec.Resources.Requests.Storage().Value()))) <= int(resource.Tera) {
+				return fmt.Errorf("calculating the current OSD size in TiB might result in overflow")
+			}
+			// Get rounded up value and convert to actual
+			currOSDSizeInTiB = float64(desiredStorageDeviceSet.DataPVCTemplate.Spec.Resources.Requests.Storage().ScaledValue(resource.Tera) - 1)
 			currDeviceSetCount = desiredStorageDeviceSet.Count
 		}
 
@@ -287,7 +298,8 @@ func (r *dataFoundationReconciler) reconcileStorageCluster() error {
 			ds = desiredStorageDeviceSet
 		}
 
-		// Prevent downscaling by comparing count from secret and count from storage cluster
+		// Prevent downscaling by comparing OSD size and count from secret and from storage cluster
+		r.setDeviceSetOSDSize(ds, desiredOSDSizeInTiB, currOSDSizeInTiB)
 		r.setDeviceSetCount(ds, desiredDeviceSetCount, currDeviceSetCount)
 		r.storageCluster.Spec = sc.Spec
 
@@ -318,6 +330,21 @@ func (r *dataFoundationReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDev
 	} else {
 		r.Log.V(-1).Info("Requested storage device set count will result in downscaling, which is not supported. Skipping")
 		deviceSet.Count = currDeviceSetCount
+	}
+}
+
+func (r *dataFoundationReconciler) setDeviceSetOSDSize(ds *ocsv1.StorageDeviceSet, desiredOSDSize float64, currOSDSize float64) {
+	r.Log.Info("Setting storage device set OSD size", "Current", currOSDSize, "New", desiredOSDSize)
+	updatedOSDSize := 0.0
+	if currOSDSize <= desiredOSDSize {
+		updatedOSDSize = desiredOSDSize
+
+	} else {
+		r.Log.V(-1).Info("Requested storage device set OSD size will result in downscaling, which is not supported. Skipping")
+		updatedOSDSize = currOSDSize
+	}
+	ds.DataPVCTemplate.Spec.Resources.Requests = corev1.ResourceList{
+		"storage": resource.MustParse(fmt.Sprintf("%vTi", updatedOSDSize)),
 	}
 }
 
--- a/controllers/datafoundation.go
+++ b/controllers/datafoundation.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"math"
-	"strconv"
 	"strings"
 	"time"
 
@@ -52,7 +51,7 @@ const (
 )
 
 type dataFoundationSpec struct {
-	UsableCapacityInTiB     int
+	UsableCapacityInTiB     resource.Quantity
 	OnboardingValidationKey string
 }
 
@@ -264,7 +263,10 @@ func (r *dataFoundationReconciler) reconcileOnboardingValidationSecret() error {
 func (r *dataFoundationReconciler) reconcileStorageCluster() error {
 	r.Log.Info("Reconciling StorageCluster")
 
-	if r.spec.UsableCapacityInTiB <= 0 {
+	UsableCapacityInTiB := r.spec.UsableCapacityInTiB
+	OSDSizeUpperBoundInTiB := templates.OSDSizeUpperBoundInTib
+
+	if UsableCapacityInTiB.CmpInt64(0) <= 0 {
 		return fmt.Errorf("invalid or missing field: UsableCapacityInTiB, provided value is %v", r.spec.UsableCapacityInTiB)
 	}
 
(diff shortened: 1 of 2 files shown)


controllers/datafoundation.go
```diff
@@ -271,25 +272,48 @@ func (r *dataFoundationReconciler) reconcileStorageCluster() error {
 			return err
 		}
 
-		// Convert the desired size to the device set count based on the underlaying OSD size
-		desiredDeviceSetCount := int(math.Ceil(float64(r.spec.UsableCapacityInTiB) / templates.OSDSizeInTiB))
+		// get the capacity in TiB planned to be consumed by user
+		storageClusterSize := math.Ceil(float64(r.spec.UsableCapacityInTiB))
+		if storageClusterSize <= templates.VerticalScalerUpperBoundInTiB {
```
rchikatw: No need to use an extra variable because it's been used only in one place.

rchikatw (commented):
Also squash your commits in a single commit.

controllers/datafoundation.go
```diff
@@ -271,25 +272,48 @@ func (r *dataFoundationReconciler) reconcileStorageCluster() error {
...
+			}
 
-		// Get the storage device set count of the current storage cluster
-		currDeviceSetCount := 0
-		if desiredStorageDeviceSet := findStorageDeviceSet(r.storageCluster.Spec.StorageDeviceSets, defaultDeviceSetName); desiredStorageDeviceSet != nil {
-			currDeviceSetCount = desiredStorageDeviceSet.Count
-		}
+			// Count remains 1
+			// todo need to confirm perf of all combos of Count and OSD Sizes (i.e. 2 --> count = 1 and OSD Size = 2 OR count = 2 and OSD Size = 1)
+			r.storageCluster.Spec = desiredStorageCluster.Spec
```
nb-ohad: We should not update the current storage cluster spec, we should update the template instance which override the spec (later in the function)

controllers/datafoundation.go
nb-ohad, line 1: A general comment about this change. 

The way I see it we need to calculate only 2 values:
```
desiredDeviceSetCount := int(
    math.Ceil(
        math.Max(float64(r.spec.UsableCapacityInTiB), VerticalScalerUpperBoundInTiB) / templates.OSDSizeInTiB
    )
)
```

and 

```
OSDSize  := math.Min(float64(r.spec.UsableCapacityInTiB), VerticalScalerUpperBoundInTiB)
```
And we only need one protection guard, to make sure the new desired size is bigger than the actual size.


This means that code changes here should be minor at best and the flow should stay similar to the original one

nb-ohad (changes requested):

controllers/datafoundation.go
raaizik: Don't you mean the denominator is a function of `UsableCapacityInTiB`, and not the constant `templates.OSDSizeInTiB`?, Meaning:

```go
desiredDeviceSetCount := int(
    math.Ceil(
        math.Max(float64(r.spec.UsableCapacityInTiB), VerticalScalerUpperBoundInTiB) / OSDSizeInTiB
    )
)
``` 
where 
```go
OSDSizeInTiB := math.Min(float64(r.spec.UsableCapacityInTiB), templates.VerticalScalerUpperBoundInTiB)

```

nb-ohad: What I am trying to say is that given that the max OSD size is 4 then for every given valid cluster size we can compute the OSD size and the OSD count (desiredDeviceSetCount):
```
OSD size = min(cluster size, 4)
OSD count = ceil(max(cluster size, 4) / 4)
```

Coming to think about it, there is no need to have `templates.VerticalScalerUpperBoundInTiB` at all as its value is always going to match `templates.OSDSizeInTiB`

nb-ohad: Looking at my last comment, I think we can even simplify the OSD count formula
``` 
ceil(max(cluster size, 4) / 4) => ceil(max(cluster size / 4, 1)) => ceil(clusterSize / 4) 
```
This means the actual formals are:
```
OSD size = min(cluster size, 4)
OSD count = ceil(cluster size / 4)
```

raaizik: I think my suggestion is readable. IMO there also needs to be a differentiation between the initialization value and the one associated with the vertical scaler, therefore:
```Go
OSDSizeInTiB                  = 1
VerticalScalerUpperBoundInTiB = 4
```

controllers/datafoundation.go
```diff
@@ -271,12 +273,21 @@ func (r *dataFoundationReconciler) reconcileStorageCluster() error {
 			return err
 		}
 
-		// Convert the desired size to the device set count based on the underlaying OSD size
-		desiredDeviceSetCount := int(math.Ceil(float64(r.spec.UsableCapacityInTiB) / templates.OSDSizeInTiB))
+		// get the desired capacity in TiB planned to be consumed by user in range 1-4TiB
+		desiredOSDSizeInTiB := math.Min(math.Ceil(float64(r.spec.UsableCapacityInTiB)), templates.VerticalScalerUpperBoundInTiB)
```
nb-ohad: Something does not make sense with 
```
desiredOSDSizeInTiB := math.Min(math.Ceil(float64(r.spec.UsableCapacityInTiB)), templates.VerticalScalerUpperBoundInTiB)
```

Why are we basing desired state on actual state (`r.spec.*`)?


Also `ceil(float(int))` will just return the `int` as a float. 
This means that the above statement can be simplified to
```suggestion
desiredOSDSizeInTiB := math.Min(
    float64(r.spec.UsableCapacityInTiB), 
    float64(templates.VerticalScalerUpperBoundInTiB),
)
```

controllers/datafoundation.go
```diff
@@ -271,12 +273,21 @@ func (r *dataFoundationReconciler) reconcileStorageCluster() error {
...
 
-		// Get the storage device set count of the current storage cluster
+		// Convert the desired size to the device set count based on the underlying OSD size
+		desiredDeviceSetCount := int(math.Ceil(float64(r.spec.UsableCapacityInTiB)) / desiredOSDSizeInTiB)
```
nb-ohad: Very similar questions here.
Why are we basing desired state on actual state (r.spec.*)?

Also, I believe there is a bug here. Converting to `int` will lose the friction not round it up. The actual statement should be
```suggestion
		desiredDeviceSetCount := math.Ceil(float64(r.spec.UsableCapacityInTiB) / desiredOSDSizeInTiB)
```

controllers/datafoundation.go
```diff
@@ -271,12 +273,21 @@ func (r *dataFoundationReconciler) reconcileStorageCluster() error {
...
+		// Get the storage device set count and OSD size of the current storage cluster
 		currDeviceSetCount := 0
+		currOSDSizeInTiB := 0.0
 		if desiredStorageDeviceSet := findStorageDeviceSet(r.storageCluster.Spec.StorageDeviceSets, defaultDeviceSetName); desiredStorageDeviceSet != nil {
+			// Avoid overflow
+			if len(strconv.Itoa(int(desiredStorageDeviceSet.DataPVCTemplate.Spec.Resources.Requests.Storage().Value()))) <= int(resource.Tera) {
+				return fmt.Errorf("calculating the current OSD size in TiB might result in overflow")
+			}
+			// Get rounded up value and convert to actual
+			currOSDSizeInTiB = float64(desiredStorageDeviceSet.DataPVCTemplate.Spec.Resources.Requests.Storage().ScaledValue(resource.Tera) - 1)
```
nb-ohad: There is no need to convert this into an int or a float. I believe that resource quantity provides methods so they can be compared.

```suggestion
currOSDSizeInTiB := desiredStorageDeviceSet.DataPVCTemplate.Spec.Resources.Requests.Storage()
```

controllers/datafoundation.go
```diff
@@ -287,7 +298,8 @@ func (r *dataFoundationReconciler) reconcileStorageCluster() error {
 			ds = desiredStorageDeviceSet
 		}
 
-		// Prevent downscaling by comparing count from secret and count from storage cluster
+		// Prevent downscaling by comparing OSD size and count from secret and from storage cluster
+		r.setDeviceSetOSDSize(ds, desiredOSDSizeInTiB, currOSDSizeInTiB)
```
nb-ohad: I don't see a point in having this separate from `setDevicesetCount`. Mainly as we need to protect against downscale as a result of these pair of parameters configured together. 

You might want to rename `setDeviceSetCount` to `setDeviceSetProperties` and have the entire logic in one place

datafoundation/templates/storagecluster.go
```diff
@@ -27,8 +27,9 @@ import (
 // StorageClusterTemplate is the template that serves as the base for the storage clsuter deployed by the operator
 
 const (
-	OSDSizeInTiB            = 4
-	backingStorageClassName = "default-ocs-storage-class"
+	OSDSizeInTiB                  = 1
+	VerticalScalerUpperBoundInTiB = 4
```
nb-ohad: I am fine with you wanting to change the name of the upper limit of allowed OSD size but

1. OSDSizeInTiB has no use anymore so we should not keep it
2. the new name `VerticalScalerUpperBoundInTiB` need to identify itself as belonging to the OSD value because it is an export of the entire `templates` package. 

maybe:
```suggestion
	OSDSizeUpperBoundInTiB = 4
```

nb-ohad (changes requested):

controllers/datafoundation.go
nb-ohad: Resolving in favor of a newer review comment.

datafoundation/templates/storagecluster.go
raaizik: > OSDSizeInTiB has no use anymore so we should not keep it

[datafoundation/templates/storagecluster.go#L143](https://github.com/red-hat-storage/managed-fusion-agent/blob/18f7d0634c6e8aa9b27e25c05ed85da1f1219bb8/datafoundation/templates/storagecluster.go#L143)
```Go
"storage": resource.MustParse(fmt.Sprintf("%dTi", OSDSizeInTiB)),
```
As I mentioned this is basically the initialization part. It can't be initialized to 4 as it was since then we won't be able to have the full range of 1-4TiB due to the downscaling check. That was the point I was making about being able to differentiate between initialization and scaling.

Edit: omitted `OSDSizeInTiB`

controllers/datafoundation.go
raaizik: I was about to

controllers/datafoundation.go
raaizik: > Why are we basing desired state on actual state (r.spec.*)?

That was my exact question, but I just followed the logic (that doesn't make sense) that existed prior to my change [controllers/datafoundation.go#L275](https://github.com/red-hat-storage/managed-fusion-agent/blob/main/controllers/datafoundation.go#L275)
```Go
desiredDeviceSetCount := int(math.Ceil(float64(r.spec.UsableCapacityInTiB) / templates.OSDSizeInTiB))
```

controllers/datafoundation.go
raaizik: I guess you meant the casting should wrap the entire expression (it is after all an integer):
```Go
desiredDeviceSetCount := int(math.Ceil(float64(r.spec.UsableCapacityInTiB) / desiredOSDSizeInTiB))
```

controllers/datafoundation.go
nb-ohad: The `Ceil` on the orignal call was around the division where it made sense that we will round up the result. 
In the new code it is just a `Ceil` on a float conversion. That is the difference

controllers/datafoundation.go
nb-ohad: Yep, that makes sense

controllers/datafoundation.go
raaizik: Note I was referring to the questionable use of `r.spec` in the old code and nothing else. In the new code it does make sense to base the desired state on the above since this is basically the flow required to update this variable, polled via the runtime object

controllers/datafoundation.go
```diff
@@ -271,12 +272,27 @@ func (r *dataFoundationReconciler) reconcileStorageCluster() error {
...
 
-		// Get the storage device set count of the current storage cluster
+		// get the desired capacity in TiB planned to be consumed by user in range 1-4TiB
+		desiredOSDSizeInTiBQuantity := OSDSizeUpperBoundInTiBQuantity
+		// Set desired to be the minimum
+		if UsableCapacityInTiBQuantity.Cmp(OSDSizeUpperBoundInTiBQuantity) <= 0 {
+			desiredOSDSizeInTiBQuantity = UsableCapacityInTiBQuantity
+		} else {
+			desiredOSDSizeInTiBQuantity = OSDSizeUpperBoundInTiBQuantity
+		}
```
rewantsoni: We don't need a else here

rewantsoni:
- Can we merge all the commits to be a single commit?
- Also, could you remove WIP from the PR title?

controllers/datafoundation.go
```diff
@@ -50,7 +51,7 @@ const (
 )
 
 type dataFoundationSpec struct {
-	UsableCapacityInTiB     int
+	UsableCapacityInTiB     resource.Quantity
```
nb-ohad: This is an external contract/api/input in the form of a CR interface. We cannot change the type here as it is expected to just be a number. lets keep this as an int (we don't have a choice)

controllers/datafoundation.go
```diff
@@ -262,7 +263,10 @@ func (r *dataFoundationReconciler) reconcileOnboardingValidationSecret() error {
 func (r *dataFoundationReconciler) reconcileStorageCluster() error {
 	r.Log.Info("Reconciling StorageCluster")
 
-	if r.spec.UsableCapacityInTiB <= 0 {
+	UsableCapacityInTiB := r.spec.UsableCapacityInTiB
```
nb-ohad: As of the above comment. we will need to translate from int to quantity only once here

controllers/datafoundation.go
```diff
@@ -262,7 +263,10 @@ func (r *dataFoundationReconciler) reconcileOnboardingValidationSecret() error {
...
+	OSDSizeUpperBoundInTiB := templates.OSDSizeUpperBoundInTib
+
+	if UsableCapacityInTiB.CmpInt64(0) <= 0 {
 		return fmt.Errorf("invalid or missing field: UsableCapacityInTiB, provided value is %v", r.spec.UsableCapacityInTiB)
 	}
```
nb-ohad: The original intention for this was a sanity check on the provided input (the datafoundation spec) and as such was the first step of the reconcile function, before anything else. 

Is there a reason to change it to work against a different variable and against a quantity and not an int (the spec is given as an int)

datafoundation/templates/storagecluster.go
```diff
@@ -27,10 +25,12 @@ import (
...
 
 const (
-	OSDSizeInTiB            = 4
-	backingStorageClassName = "default-ocs-storage-class"
+	OSDSizeUpperBoundInTiBUnscaled = 4398046511104
+	backingStorageClassName        = "default-ocs-storage-class"
 )
 
+var OSDSizeUpperBoundInTib = resource.NewQuantity(OSDSizeUpperBoundInTiBUnscaled, resource.BinarySI)
+
```
nb-ohad: If we cannot do an actual cost, can we mitigate the chance that someone will change the var value from the outside by using a functions?

```suggestion
	backingStorageClassName        = "default-ocs-storage-class"
)

func GetOSDSizeUpperBoundInTib() resource.Quantity {
    return resource.NewQuantity(4398046511104, resource.BinarySI)
}
```

nb-ohad (changes requested):

datafoundation/templates/storagecluster.go
raaizik: Switched back to `MustParse` in function scope

controllers/datafoundation.go
```diff
@@ -271,24 +272,34 @@ func (r *dataFoundationReconciler) reconcileStorageCluster() error {
...
-			currDeviceSetCount = desiredStorageDeviceSet.Count
+		// Get the desired capacity in TiB planned to be consumed by user in range 1-4TiB
+		// Set desired to be the minimum
+		desiredOSDSizeInTiB := UsableCapacityInTiB
+		if OSDSizeUpperBoundInTiB.Cmp(UsableCapacityInTiB) < 0 {
+			desiredOSDSizeInTiB = OSDSizeUpperBoundInTiB
 		}
 
+		// Convert the desired size to the device set count based on the underlying OSD size
+		desiredDeviceSetCount := int(math.Ceil(UsableCapacityInTiB.AsApproximateFloat64() / desiredOSDSizeInTiB.AsApproximateFloat64()))
```
nb-ohad: No need to transform back to bytes and lose precision, you can use `ScaledValue` as a better alternative.
```suggestion
		desiredDeviceSetCount := int(math.Ceil(
			float(UsableCapacityInTiB.ScaledValue("Tera")) / float(desiredOSDSizeInTiB.ScaledValue("Tera"))
		))
```

nb-ohad: @raaizik This is not a must, if you feel the code is good as is we can merge it as is

raaizik: @nb-ohad At some point in the future we would probably need to support non-tera values, no?

nb-ohad: Storage provisioning for an entire cluster (not a volume provided by the cluster) is provisioned at the minimum GiBs levels, but in most cases at the TiB level. I would even make a case that our code should prevent allowing clusters to be provisioned at a unit below the GiB unit. 

If you feel that you prefer to translate everything to GiBs, I am ok with that. The main thing is that I would like to eliminate the imprecisions that come with translating to very big floats (in this case, translating to bytes)

